PR #32, opened 2013-01-14
New IPv4 filter implementation that isn't a memory hog -- fixes rtorrent issue #106 and #71

sallyswiss:
New extents.h data structure in libtorrent so that it does not waste memory.  it is now an ordered map instead of a crazy sparse hash, much better, still log n access time.  it also stores ranges as in p2p file format so it can handle p2p data from bluetack lists as well as cidr notation which is now only thing supported. There is corresponding commit and pull request for rtorrent that makes the necessary changes there too.

### DIFF
--- a/src/torrent/peer/peer_list.cc
+++ b/src/torrent/peer/peer_list.cc
@@ -44,6 +44,8 @@
 #include "download/available_list.h"
 #include "torrent/peer/client_list.h"
 
+#include "torrent/utils/log.h"
+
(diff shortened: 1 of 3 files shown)


chros73:
Rebased on current master branch in: https://github.com/rakshasa/libtorrent/pull/162

chros73:
This can be closed since the other one is merged: https://github.com/rakshasa/libtorrent/pull/162

PR #44, opened 2013-10-04
Fix string_manip.h

pyhalov:
copy_escape_html() shouldn't depend on locale. It should encode
every string which is not ascii digit or alphabetic symbol.

In particular, this doesn' work on  OpenIndiana with g++ 4.7.3.

### DIFF
--- a/rak/string_manip.h
+++ b/rak/string_manip.h
@@ -190,9 +190,10 @@ template <typename InputIterator, typename OutputIterator>
 OutputIterator
 copy_escape_html(InputIterator first, InputIterator last, OutputIterator dest) {
   while (first != last) {
-    if (std::isalpha(*first, std::locale::classic()) ||
-        std::isdigit(*first, std::locale::classic()) ||
-        *first == '-') {
+    if (isascii(*first)&&
+         (std::isalpha(*first, std::locale::classic()) ||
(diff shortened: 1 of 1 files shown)


rakshasa:
```
Obtains a reference to the C++ locale that implements the classic "C" locale semantics. This locale, unlike the global locale, cannot be altered.
```

OpenIndiana implements std::locale::classic() wrong, and as such should keep a separate patch set to handle these cases.

pyhalov:
illumos devs says that it's broken libtorrent logic  - https://www.illumos.org/issues/4412

risto3:
> OpenIndiana implements std::locale::classic() wrong, and as such should keep a separate patch set to handle these cases.
?? I believe this is implemented rather by gcc in libstdc++-v3/src/c++98

pyhalov:
It's not OI, it's gcc. Related gcc bug report is https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58628, corresponding illumos bug report https://www.illumos.org/issues/4173, discussion on illumos-dev - http://comments.gmane.org/gmane.os.illumos.devel/14193

rakshasa:
The patch does an 'isascii' check, while classic locale is always supposed to be "C" ASCII locale.

So not a but in libtorrent.

PR #78, opened 2015-03-01
Removes the tr1 prefix from C++11 libraries

zachberger:


rakshasa:
What exactly is the reason for this patch?

The current macros are used for compatibility reasons when compiling C++11 on gcc.

zachberger:
My apologies. Last time I had tried to compile with `clang` on OS X 10.10 tr1 libraries were not found. Doesn't seem to be a problem now

rakshasa:
This might be an issue with the configure.ac script not detecting that you're supposed to be compiling without the tr1 prefix.

```
libtorrent>cat scripts/rak_cxx.m4
...

AC_DEFUN([RAK_CHECK_TR1_LIB], [
```

I don't know why your environment would define the versions with the prefix, however I'd be happy to include any patch to fix this autoconf script.

Issue #84, opened 2015-07-11
Sort files by path for a more compact and nicer directory tree


mrvn:
Every now and then I get a huge collection torrent where the files are listed in random order, sorted by size or worst with BitComet padding files. Often I only want part of the collection but selecting all files from just one directory a huge pain since they are spread out all over the listing. Files should be sorted so all files from the same subdir appear together.

PR #85, opened 2015-07-11
Sort files by path for a more compact and nicer directory tree

mrvn:
This patch tries to implement issue #84 "Sort files by path for a more compact and nicer directory tree".

Unfortunately it makes FileList::create_chunk() and FileList::inc_completed() more expensive. FileList::create_chunk() becomes O(n^2) [was O(n)] where n is the number of files. But the worst case only happens if files are in reverse order in the torrent and verry small. In general it is linear with the number of files. FileList::inc_completed() becomes O(n) [was amortized O(1 + n/num_chunks)].

Can't say I noticed a difference in cpu load in practice but I don't have a torrent with many files, small chunks and high download rate to compare at the moment.

### DIFF
--- a/src/download/download_constructor.cc
+++ b/src/download/download_constructor.cc
@@ -331,6 +331,7 @@ DownloadConstructor::parse_multi_files(const Object& b, uint32_t chunkSize) {
 
   fileList->initialize(torrentSize, chunkSize);
   fileList->split(fileList->begin(), &*splitList.begin(), &*splitList.end());
+  fileList->sort();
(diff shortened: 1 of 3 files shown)


rakshasa:
This is really a display issue, and doing this in libtorrent would make future changes involving file lists more difficult.

It should be cheap enough to do the sorting and culling in rtorrent instead.

rakshasa:
Closing this as it's the wrong approach. You can't change the order of the files in libtorrent, it needs to be client-side.

PR #86, opened 2015-07-12
Update configure.ac

robshelc:


robshelc:
Closing this as I pointed it to master. Looks like rakshasa:branch-0.13 already has this change.

Issue #98, opened 2015-11-30
ProtocolExtension::send_metadata_piece build_bencode buffer too small on x64, remote user can cause internal_error


X-cela:
Requesting piece -1 during a meta extension req  from a x64 client allows remote user to crash application 

```
ProtocolExtension::send_metadata_piece(size_t piece) 
...
  if (m_download->info()->is_meta_download() || piece >= pieceEnd) {
    // reject: { "msg_type" => 2, "piece" => ... }
    m_pendingType = UT_METADATA;
    m_pending = build_bencode(40, "d8:msg_typei2e5:piecei%zuee", piece);
    return;
  }
```

40 is too small for 64bit size_t piece

```
torrent::ProtocolExtension::send_metadata_piece (this=0x171c2e0, piece=18446744073709551615) at extensions.cc:
392       size_t pieceEnd = (metadataSize + metadata_piece_size - 1) >> metadata_piece_shift;
(gdb) p/x piece
$1 = 0xffffffffffffffff
(gdb) n
397         m_pending = build_bencode(40, "d8:msg_typei2e5:piecei%zuee", piece);

torrent::ProtocolExtension::build_bencode (maxLength=40, format=0x7feb89071b71 "d8:msg_typei2e5:piecei%zuee") at extensions.cc:170
174       unsigned int length = vsnprintf(b, maxLength, format, args);
(gdb) 
177       if (length > maxLength)
                 throw internal_error("ProtocolExtension::build_bencode wrote past buffer.");
(gdb) p length
$2213 = 44
```

```
$ ./rtorrent 
Caught internal_error: ProtocolExtension::build_bencode wrote past buffer.
/lib/libtorrent.so.19(_ZN7torrent14internal_error10initializeERKSs+0x6d) [0x7feb88f5b281]
./rtorrent(_ZN7torrent14internal_errorC1EPKc+0x7e) [0x499b10]
/lib/libtorrent.so.19(+0x22b111) [0x7feb89039111]
/lib/libtorrent.so.19(+0x22ab5a) [0x7feb89038b5a]
/lib/libtorrent.so.19(+0x22a99a) [0x7feb8903899a]
/lib/libtorrent.so.19(+0x22a310) [0x7feb89038310]
/lib/libtorrent.so.19(+0x22d815) [0x7feb8903b815]
/lib/libtorrent.so.19(+0x22e301) [0x7feb8903c301]
[...]

Issue #89, opened 2015-09-20
Not handling internal_error in event_read allows remote user to crash


X-cela:
Remote user can send handshake that will cause internal_error to be thrown and not caught causing rtorrent to crash 

handshake.cc
````C
case READ_ENC_IA:
      // Just read (and automatically decrypt) the initial payload
      // and leave it in the buffer for READ_INFO later.
      if (m_encryption.length_ia() > 0 && !fill_read_buffer(m_encryption.length_ia()))
        break;

      if (m_readBuffer.remaining() > m_encryption.length_ia())
        throw internal_error("Read past initial payload after incoming encrypted handshake.");
```

Caught internal_error: Read past initial payload after incoming encrypted handshake.
libtorrent.so.19(_ZN7torrent14internal_error10initializeERKSs+0x6d) [0x7f1b6322d281]
./rtorrent(_ZN7torrent14internal_errorC1EPKc+0x7e) [0x499b10]
libtorrent.so.19(+0x22debe) [0x7f1b6330debe]
libtorrent.so.19(_ZN7torrent9PollEPoll7performEv+0x1c6) [0x7f1b6323a7e4]
libtorrent.so.19(_ZN7torrent9PollEPoll7do_pollEli+0x1a6) [0x7f1b6323aa1c]
libtorrent.so.19(_ZN7torrent11thread_base10event_loopEPS0_+0x2a7) [0x7f1b632abe8b]
./rtorrent() [0x43968b]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf5) [0x7f1b61cdcec5]
./rtorrent() [0x435db9]

PR #99, opened 2015-12-01
Fixes to build_bencode callers, READ_ENC_IA, and object_read_bencode_c

X-cela:
Calls into build_benocde that use %zu could crash on 64 bit machines due to the size change of size_t 

Someone can force READ_ENC_IA to fail allowing an internal_error to be thrown and bring down the client, throw handshake_error instead 
#98 #89


timofonic:
Any news about this?

It seems libtorrent didn't have updates for some months.

Are there plans to make this more of a community project? I understand rakshasa is a very capable developer, but maybe some more collaboratively development could be more effective with proper management ;)

X-cela:
Added fix for #100 

I sent an email to Rakshasa asking to accept my pull request, but didn't get a response. Not sure what that means for the future of libtorrent.

PR #109, opened 2016-08-23
Fix honoring system.file.allocate.set=1 rtorrent config setting

chros73:
Fix honoring `system.file.allocate.set=1` rtorrent config setting:
- allocate space for those files only that don't have priority set to `off`.

It's useful with the following watch directory entry that only loads (in `closed` state) but not starts torrents, so priority of files can be set.
`schedule2 = watch_dir_5,  7, 10, "load.normal=/path/to/watch-dir/load_but_not_start/*.torrent"`

Fixes: https://github.com/rakshasa/rtorrent/issues/488


chros73:
This one really bothers me.
Actually I started to work on this issue for about 2 months ago (again) and this one triggered the other 4 :)

I completely understand why you didn't want to merge this in its current state, @rakshasa .
Let's modify it that it should never be able to crash the client:
- not when a download is started (this part is finished)
- not when priority if files are changed

The latter gives me major headache :)

chros73:
Omg, what happened! :) I'll create a clean pull request for this.

PR #116, opened 2016-10-21
first draft of async DNS resolution, via udns

slingamn:
This isn't production-quality code, but I wanted to check in with you before going further.

This is aimed at addressing the DNS-related cause of rakshasa/rtorrent#180. It removes the synchronous `getaddrinfo` call in the UDP tracker implementation, replacing it with async resolution via the [udns](http://www.corpit.ru/mjt/udns.html) library.

I chose udns over c-ares and adns mainly because it seemed much easier to integrate into libtorrent's existing event loop --- it only uses a single UDP socket, which was a natural fit with libtorrent's APIs. One downside is that it's not capable of failing over to TCP. Quoting the manpage: "[udns] uses UDP only, but uses EDNS0 (RFC2671) extensions which allows larger UDP buffers." This may make it incompatible with certain old DNS servers.

Anyway, I'd love to get your feedback. I can change the approach, or I can work on rebasing this against your latest ipv6 work.


configure.ac
```diff
@@ -19,7 +19,6 @@ AC_SUBST(LIBTORRENT_INTERFACE_VERSION_NO)
 
 AM_INIT_AUTOMAKE
 AC_CONFIG_HEADERS(config.h)
-AM_PATH_CPPUNIT(1.9.6)
```
slingamn: I was affected by this issue: https://github.com/OpenLightingProject/ola/issues/1043

rakshasa:
I'm open to this being potentially merged in the future, though it should be an optional dependency.

The use of a resolver in connection_manager was for the purpose of allowing different dns resolvers to be used, and so the code really needs to respect that.

slingamn:
Cool!

I wanted to clarify the ideas of "optional dependency" and of multiple DNS resolvers. As I understand it, rakshasa/rtorrent#180 is considered a valid bug, that is to say, blocking the main thread for a resolve is undesired behavior and the goal is to stop it from happening. So if we're talking about supporting different resolvers, it seems like the old synchronous `getaddrinfo`-based resolver shouldn't be one of them.

Are you considering supporting c-ares or adns? If so, what's the reason not to use udns always instead, adding a hard dependency on udns? (Concerns about its availability on different platforms, or the TCP compatibility issue, or code size, or something else?)

In passing, I also had difficulty understanding the meaning of "blocked" here:

https://github.com/rakshasa/libtorrent/blob/master/src/torrent/connection_manager.h#L158

Thanks for your time.

rakshasa:
It's not really a bug, as much as it is the lack of a feature. There's too many weird setups on which rtorrent is run that I hesitate to force dependencies when not required.

The 'blocked' comment is obsolete, and was written when sigc++ was used. Now instead the caller receives a pointer to the slot that they can bind to an empty std::function.

slingamn:
This rationale for making udns an optional dependency makes sense.

If I'm understanding correctly, the "slot"-based API will not work correctly with udns (and probably not with other asynchronous DNS libraries). The problem is that udns is pure C and isn't aware of function objects: instead, it takes callbacks in the form of a function pointer and a `void*` containing additional data. Right now, I'm using a compatibility shim where the function pointer is to a wrapper function and the `void*` points to the C++ function object that is the actual callback. But since udns is only making a shallow copy of the function object, if the original storage for the function object is deallocated (e.g., if the `TrackerUdp` object is destroyed), then any leftover callbacks that trigger will cause a use-after-free. The safe thing is not to overwrite the callback with an empty `std::function`, but to use the `dns_cancel` function to remove the callback and prevent it from being called (as in the current PR).

My current plan is to:

1. Rebase against the latest master
1. Modify the resolver API to work with `dns_cancel`
1. Use `#ifdef`s to ensure that libtorrent can be built without udns, in which case it will resolve all queries synchronously

Thoughts?

slingamn:
Closing in favor of #133.

PR #118, opened 2016-10-23
Use pkg-config for cppunit.

rakshasa:


Pro-pra:
you need also fix test/configure.ac

PR #122, opened 2016-10-31
Calculate coverage for coveralls.io

klusark:
If you want to support this, you just need to add the project on https://coveralls.io. Coverage should help with new test creation as it's hard to see what tests to write without coverage data.


klusark:
You can view what it looks like for my fork here https://coveralls.io/github/klusark/libtorrent

rakshasa:
Could you add this to rtorrent too?

Issue #130, opened 2016-11-07
1byte Heap Overead in object_read_bencode_c_string


RealAlexJones:
There is a 1byte over read in object_read_bencode_c_string after it reads in the length and is doing checks on the length

During the handshake the peer sends the following bencode as the extension handshake
bencode:
```
d4294967295
```

```
==29695==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000009b7b at pc 0x7f25ade0d91c bp 0x7ffc50962260 sp 0x7ffc50962250
READ of size 1 at 0x602000009b7b thread T0 (rtorrent main)
    #0 0x7f25ade0d91b in torrent::object_read_bencode_c_string(char const*, char const*) /libtorrent/src/torrent/object_stream.cc:107
    #1 0x7f25ade12bbb in torrent::static_map_read_bencode_c(char const*, char const*, torrent::static_map_entry_type*, torrent::static_map_mapping_type const*, torrent::static_map_mapping_type const*) /libtorrent/src/torrent/object_stream.cc:681
    #2 0x7f25adf64e06 in static_map_read_bencode<torrent::ext_handshake_keys, 7ul> ../torrent/object_stream.h:97
    #3 0x7f25adf64e06 in torrent::ProtocolExtension::parse_handshake() /libtorrent/src/protocol/extensions.cc:283
    #4 0x7f25adf667ef in torrent::ProtocolExtension::read_done() /libtorrent/src/protocol/extensions.cc:247
    #5 0x7f25adf7e365 in torrent::PeerConnectionBase::down_extension() /libtorrent/src/protocol/peer_connection_base.cc:725
    #6 0x7f25adf9b345 in torrent::PeerConnection<(torrent::Download::ConnectionType)1>::read_message() /libtorrent/src/protocol/peer_connection_leech.cc:347
    #7 0x7f25adf9b345 in torrent::PeerConnection<(torrent::Download::ConnectionType)1>::event_read() /libtorrent/src/protocol/peer_connection_leech.cc:400
    #8 0x7f25adf7583a in torrent::HandshakeManager::receive_succeeded(torrent::Handshake*) /libtorrent/src/protocol/handshake_manager.cc:225
    #9 0x7f25adf6d667 in torrent::Handshake::event_write() /libtorrent/src/protocol/handshake.cc:966
    #10 0x7f25ade210a5 in torrent::PollEPoll::perform() /libtorrent/src/torrent/poll_epoll.cc:190
[...]

Issue #129, opened 2016-11-05
Unhandled bencode_error in TrackerHttp::process_success key "peers6"


RealAlexJones:
TrackerHttp::process_success does not handle bencode errors for key peers6 like it does for key peers

```

  if (object.has_key("peers")) {
    try {
      // Due to some trackers sending the wrong type when no peers are
      // available, don't bork on it.
      if (object.get_key("peers").is_string())
        l.parse_address_compact(object.get_key_string("peers"));

      else if (object.get_key("peers").is_list())
        l.parse_address_normal(object.get_key_list("peers"));

    } catch (bencode_error& e) {
      return receive_failed(e.what());
    }
  }

  if (object.has_key("peers6"))
    l.parse_address_compact_ipv6(object.get_key_string("peers6"));
```
If a tracker response has the key peers6, but it doesn't have a string value a bencode_error is thrown and not handled. 

Tracker Response:
```
d6:peers6i1ee
```

```
$ rtorrent
rtorrent: Wrong object type.
```
try catch is needed or the use of has_key_string

Issue #126, opened 2016-11-03
object_read_string uncaught exception from str.resize on x86


RealAlexJones:
object_read_string reads in the length of the string via >> and resizes str accordingly 

```
bool
object_read_string(std::istream* input, std::string& str) {
  uint32_t size;
  *input >> size;

  if (input->fail() || input->get() != ':')
    return false;
  
  str.resize(size);

...
}
```

resize throws an exception if size > than max_size. max_size depends on architecture and possibly implementation. On a x86 ubuntu machine max_size is 0x3ffffffc.  Having a key length of -1 will attempt to resize if 0xffffffff causing rTorrent to close 

Not all bencode parsing paths use object_read_string. Reading a .torrent file and parsing tracker messages do 

Tracker returns
```
d-1:Ai1ee
```

```
$ rtorrent
rtorrent: basic_string::resize
```

```
Breakpoint 1, 0xb7551400 in std::string::resize(unsigned int, char)@plt ()
   from /usr/lib/i386-linux-gnu/libtorrent.so.14
(gdb) bt
#0  0xb7551400 in std::string::resize(unsigned int, char)@plt ()
   from /usr/lib/i386-linux-gnu/libtorrent.so.14
#1  0xb756710d in ?? () from /usr/lib/i386-linux-gnu/libtorrent.so.14
#2  0xb756b849 in torrent::object_read_bencode(std::istream*, torrent::Object*, unsigned int) () from /usr/lib/i386-linux-gnu/libtorrent.so.14
#3  0xb756d02c in torrent::operator>>(std::istream&, torrent::Object&) ()
   from /usr/lib/i386-linux-gnu/libtorrent.so.14
#4  0xb760727e in ?? () from /usr/lib/i386-linux-gnu/libtorrent.so.14
#5  0xb7607c32 in ?? () from /usr/lib/i386-linux-gnu/libtorrent.so.14
#6  0xb755b1e0 in std::tr1::function<void ()>::operator()() const ()
   from /usr/lib/i386-linux-gnu/libtorrent.so.14
#7  0xb755af98 in torrent::Http::trigger_done() ()
   from /usr/lib/i386-linux-gnu/libtorrent.so.14
#8  0x081173aa in ?? ()
#9  0x081178a3 in ?? ()
#10 0x0811a74e in ?? ()
#11 0xb756ecd0 in torrent::PollEPoll::perform() ()
   from /usr/lib/i386-linux-gnu/libtorrent.so.14
[...]

Issue #125, opened 2016-11-03
Floating Point Exception during metadownload with  0 metadata_size


RealAlexJones:
When the client reaches out to a peer for a metadownload the peer specifies the size of the metadata via the key metadata_size in the extension handshake. 

libTorrent sets the m_chunkSize == metadata_size via FileList::reset_filesize

m_chunkSize is later used on the right hand side of modulo causing a division by 0

```
Program received signal SIGFPE, Arithmetic exception.
0x00007f0806e6a3d6 in torrent::FileList::chunk_index_size (this=0x61a000016a88, index=<optimized out>) at file_list.cc:169
169       if (index + 1 != size_chunks() || size_bytes() % chunk_size() == 0)
```

Setting metadata size to 0 in extension handshake
```
d1:md11:ut_metadatai3ee13:metadata_sizei0ee
```

```
Caught Floating point exception, dumping stack:
/usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/libasan.so.1(backtrace+0x49) [0x7f9eb905daf9]
rtorrent() [0x42d6c9]
rtorrent() [0x436e60]
rtorrent() [0x4ade65]
/lib64/libpthread.so.0(+0x10d70) [0x7f9eb85c0d70]
/usr/local/lib64/libtorrent.so.19(_ZNK7torrent8FileList16chunk_index_sizeEj+0xb6) [0x7f9eba8b73d6]
/usr/local/lib64/libtorrent.so.19(_ZNSt5_BindIFSt7_Mem_fnIMN7torrent8FileListEKFjjEEPS2_St12_PlaceholderILi1EEEE6__callIjIOjEILm0ELm1EEEET_OSt5tupleIIDpT0_EESt12_Index_tupleIIXspT1_EEE+0x65) [0x7f9eba992355]
/usr/local/lib64/libtorrent.so.19(_ZNSt5_BindIFSt7_Mem_fnIMN7torrent8FileListEKFjjEEPS2_St12_PlaceholderILi1EEEEclIJjEjEET0_DpOT_+0x40) [0x7f9eba9923c0]
/usr/local/lib64/libtorrent.so.19(_ZNSt17_Function_handlerIFjjESt5_BindIFSt7_Mem_fnIMN7torrent8FileListEKFjjEEPS4_St12_PlaceholderILi1EEEEE9_M_invokeERKSt9_Any_dataj+0x2d) [0x7f9eba99240d]
[...]

PR #131, opened 2016-11-07
Bug Fixes for 123,124,125,126,129 and 130

### LINKED ISSUE #124: internal error during metadownload if peer set metadata_size too large
During a meta download a peer specifies the length of the metadata via the key metadata_size. A peer can send a very large (or negative) number as the metadata_size. When the client attempts to check how many bytes it has left for this file an internal error is thrown. Attempting to remove the torrent will cause FileList::bytes_left to be called. 

Extension Handshake Setting metadata_size to -1:
```
d1:md11:ut_metadatai3ee13:metadata_sizei-1ee
```

```
FileList::left_bytes
...
  uint64_t left = size_bytes() - completed_bytes();
  if (left > ((uint64_t)1 << 60))
    throw internal_error("FileList::bytes_left() is too large.", data()->hash());
...
```
```
(gdb) p size_bytes()
$1 = 107339822789768
```
```
Caught internal_error: FileList::bytes_left() is too large. [#FB88BA396DD408BCE51F41AEAA71BE695F60FA87]
/usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/libasan.so.1(backtrace+0x49) [0x7effcff5daf9]
/usr/local/lib64/libtorrent.so.19(_ZN7torrent14internal_error10initializeERKSs+0xd4) [0x7effd177b074]
/usr/local/lib64/libtorrent.so.19(_ZN7torrent14internal_errorC2EPKcRKNS_10HashStringE+0xeb) [0x7effd17bdf3b]
/usr/local/lib64/libtorrent.so.19(_ZNK7torrent8FileList10left_bytesEv+0xef) [0x7effd17b7aaf]
/usr/local/lib64/libtorrent.so.19(_ZNSt5_BindIFSt7_Mem_fnIMN7torrent8FileListEKFmvEEPS2_EE6__callImIEILm0EEEET_OSt5tupleIIDpT0_EESt12_Index_tupleIIXspT1_EEE+0x40) [0x7effd1892580]
/usr/local/lib64/libtorrent.so.19(_ZNSt5_BindIFSt7_Mem_fnIMN7torrent8FileListEKFmvEEPS2_EEclIIEmEET0_DpOT_+0x20) [0x7effd18925c0]
/usr/local/lib64/libtorrent.so.19(+0x29a710) [0x7effd18ec710]
/usr/local/lib64/libtorrent.so.19(_ZN7torrent11TrackerList10send_stateEPNS_7TrackerEi+0x15c) [0x7effd179ec7c]
/usr/local/lib64/libtorrent.so.19(_ZN7torrent17TrackerController15send_stop_eventEv+0x293) [0x7effd179c893]
/usr/local/lib64/libtorrent.so.19(_ZN7torrent8Download4stopEi+0x1a4) [0x7effd17750f4]
rtorrent() [0x553acb]
rtorrent() [0x49363d]
rtorrent() [0x495362]
rtorrent() [0x4953ea]
rtorrent() [0x495456]
rtorrent() [0x5aae95]
rtorrent() [0x5aaf14]
rtorrent() [0x5aeb38]
rtorrent() [0x5b072f]
rtorrent() [0x5c450f]
```

RealAlexJones:
Fixes for #123 #124 #125 #126 #129 #130 

One that I'm not too sure about is the change to left_bytes where I remove the restriction of left being < 1 << 60. This is to fix #124 where a peer during a meta download can set the size to be very large. I'm not sure which case that restriction was attempting to prevent.

### DIFF
--- a/src/dht/dht_server.cc
+++ b/src/dht/dht_server.cc
@@ -743,6 +743,11 @@ DhtServer::event_read() {
       if (!message[key_t].is_raw_string())
         throw dht_error(dht_error_protocol, "No transaction ID");
 
+	  //Check transaction id length to avoid use of it in dht_error if invalid
+	  if(message[key_t].as_string().size() > 20) {
+		  throw network_error();
+	  }
+
       if (!message[key_y].is_raw_string())
         throw dht_error(dht_error_protocol, "No message type");
 
--- a/src/dht/dht_server.cc
+++ b/src/dht/dht_server.cc
@@ -597,7 +597,7 @@ void
 DhtServer::create_error(const DhtMessage& req, const rak::socket_address* sa, int num, const char* msg) {
   DhtMessage error;
 
-  if (req[key_t].is_raw_bencode() || req[key_t].is_raw_string())
+  if ((req[key_t].is_raw_bencode() || req[key_t].is_raw_string()) && req[key_t].as_raw_string().size < 20)
     error[key_t] = req[key_t];
 
   error[key_y] = raw_bencode::from_c_str("1:e");
@@ -743,10 +743,10 @@ DhtServer::event_read() {
       if (!message[key_t].is_raw_string())
         throw dht_error(dht_error_protocol, "No transaction ID");
 
-	  //Check transaction id length to avoid use of it in dht_error if invalid
-	  if(message[key_t].as_string().size() > 20) {
-		  throw network_error();
-	  }
+      // Restrict the length of Transaction IDs. We echo them in our replies.
+      if(message[key_t].as_raw_string().size() > 20) {
+		  throw dht_error(dht_error_protocol, "Transaction ID length too long");
+      }
 
       if (!message[key_y].is_raw_string())
         throw dht_error(dht_error_protocol, "No message type");
--- a/src/dht/dht_server.cc
+++ b/src/dht/dht_server.cc
@@ -597,8 +597,7 @@ void
 DhtServer::create_error(const DhtMessage& req, const rak::socket_address* sa, int num, const char* msg) {
   DhtMessage error;
 
-  if ((req[key_t].is_raw_bencode() && req[key_t].as_raw_bencode().size() < 20) ||
-		  (req[key_t].is_raw_string() && req[key_t].as_raw_string().size() < 20))
+  if (req[key_t].is_raw_string() && req[key_t].as_raw_string().size() < 67)
     error[key_t] = req[key_t];
 
   error[key_y] = raw_bencode::from_c_str("1:e");
--- a/src/dht/dht_server.cc
+++ b/src/dht/dht_server.cc
@@ -597,7 +597,7 @@ void
 DhtServer::create_error(const DhtMessage& req, const rak::socket_address* sa, int num, const char* msg) {
   DhtMessage error;
 
-  if ((req[key_t].is_raw_bencode() || req[key_t].is_raw_string()) && req[key_t].as_raw_string().size < 20)
+  if ((req[key_t].is_raw_bencode() || req[key_t].is_raw_string()) && req[key_t].as_raw_string().size() < 20)
     error[key_t] = req[key_t];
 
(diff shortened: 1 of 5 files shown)


src/dht/dht_server.cc
```diff
@@ -743,6 +743,11 @@ DhtServer::event_read() {
       if (!message[key_t].is_raw_string())
         throw dht_error(dht_error_protocol, "No transaction ID");
 
+	  //Check transaction id length to avoid use of it in dht_error if invalid
+	  if(message[key_t].as_string().size() > 20) {
+		  throw network_error();
+	  }
```
rakshasa: Please try to follow the code style, e.g. use spaces, not tabs. Properly format the comment with a leading space and punctuation.

Use 'as_raw_string().size()' as that doesn't require a conversion to std::string.

Also use dht_error_protocol, not network_error.

src/torrent/data/file_list.cc
```diff
@@ -155,9 +155,6 @@ uint64_t
 FileList::left_bytes() const {
   uint64_t left = size_bytes() - completed_bytes();
 
-  if (left > ((uint64_t)1 << 60))
-    throw internal_error("FileList::bytes_left() is too large.", data()->hash());
-
```
rakshasa: You are treating the symptom, not the disease. The above check is there to catch issues in other parts of the code, notably that we don't have a proper check in 'DownloadMain::set_metadata_size' that checks if the size is zero or more than a reasonable size.

In the case of metadata sizes, (1 << 26) or 64Mb seems a good upper limit.

src/dht/dht_server.cc
rakshasa: In that case also add the length check to 'create_error' so that it doesn't try to copy the key.

src/dht/dht_server.cc
```diff
@@ -597,7 +597,7 @@ void
 DhtServer::create_error(const DhtMessage& req, const rak::socket_address* sa, int num, const char* msg) {
   DhtMessage error;
 
-  if (req[key_t].is_raw_bencode() || req[key_t].is_raw_string())
+  if ((req[key_t].is_raw_bencode() || req[key_t].is_raw_string()) && req[key_t].as_raw_string().size < 20)
```
rakshasa: The parenthesis placement is wrong, you can't call as_raw_string on is_raw_bencode.

RealAlexJones: is there ever a case where key_t will be raw_bencode and not a raw_string? 

The key is defined as t*S which will only ever be populated with raw_string

src/dht/dht_server.cc
```diff
@@ -597,7 +597,8 @@ void
 DhtServer::create_error(const DhtMessage& req, const rak::socket_address* sa, int num, const char* msg) {
   DhtMessage error;
 
-  if ((req[key_t].is_raw_bencode() || req[key_t].is_raw_string()) && req[key_t].as_raw_string().size() < 20)
+  if ((req[key_t].is_raw_bencode() && req[key_t].as_raw_bencode().size() < 20) ||
```
rakshasa: If the key is 20 bytes long string, then the bencode version of it will be '20:<20*bytes>" which is 23 bytes.

However the key can be more than 20 char long strings, so perhaps we should use e.g. 64 and 67 as the max lengths.

src/dht/dht_server.cc
```diff
@@ -597,7 +597,8 @@ void
...
+		  (req[key_t].is_raw_string() && req[key_t].as_raw_string().size() < 20))
```
rakshasa: If the key is 20 bytes long string, then the bencode version of it will be '20:<20*bytes>" which is 23 bytes.

However the key can be more than 20 char long strings, so perhaps we should use e.g. 64 and 67 as the max lengths.

rakshasa: The internets is somewhat ambiguous about what the key can be, so I guess we can make the assumption that it's always a raw_string since the rest of the code assumes it.

PR #133, opened 2016-11-07
second draft of async DNS resolution via udns

slingamn:
This is the successor to #116.

The main things that have changed:

1. ipv6 support (i.e., support for both A and AAAA queries, in parallel if necessary)
1. support for building without udns
1. a change to the async resolution API (unfortunately adding some complexity)

The new API is because the previous version of this patch had a bug:

````
Caught internal_error: priority_queue_insert(...) called on an already queued item.
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(_ZN7torrent14internal_error10initializeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x243) [0xb75b8293]
/home/shivaram/workspace/sandbox/bin/rtorrent(_ZN7torrent14internal_errorC2EPKc+0x86) [0x808fe46]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(+0xeab48) [0xb767cb48]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(+0x82f7a) [0xb7614f7a]
/lib/libudns.so.0(+0x340b) [0xb756e40b]
/lib/libudns.so.0(dns_ioevent+0x5ad) [0xb75705dd]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(_ZN7torrent9UdnsEvent10event_readEv+0x1d) [0xb7614ecd]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(_ZN7torrent9PollEPoll7performEv+0x16e) [0xb75cfbfe]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(_ZN7torrent9PollEPoll7do_pollExi+0x67) [0xb75cfca7]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.19(_ZN7torrent11thread_base10event_loopEPS0_+0x1a2) [0xb7614212]
/home/shivaram/workspace/sandbox/bin/rtorrent() [0x80549fe]
/lib/libc.so.6(__libc_start_main+0xf6) [0xb6faa5a6]
/home/shivaram/workspace/sandbox/bin/rtorrent() [0x8055b47]
````

the root cause of which was my attempt to combine the `dns_submit_a4` and `dns_timeouts` operations into one `resolve_in4` operation. Basically:

1. `dns_timeouts` can call arbitrary callbacks
2. The callback `TrackerUdp::start_announce` can call into `TrackerUdp::receive_failed`, and then into `TrackerList::receive_failed` and ultimately into `TrackerController::do_timeout`, which can then make *new* DNS queries via `TrackerUdp::send_state`
3. Under rare circumstances (when a server name becomes resolvable again after a period of failure), this can result in multiple calls to `TrackerUdp::send_state` happening on the same tracker, one on top of the other on the call stack
4. Since in #116, the query identifier doesn't get returned to the client until after `dns_timeouts` finishes executing all callbacks, `TrackerUdp::close_directly` was unable to cancel the redundant query (because it wasn't stored in `m_udns_query` yet)
5. which could finally result in two overlapping calls to `TrackerUdp::start_announce` and the above assertion failure

I came to the conclusion that the only way to make the API safe for re-entry in general is to immediately return the query identifier to the client, before calling any callbacks --- hence splitting the resolve operation into "queue" and "flush" operations, corresponding to the `dns_submit_a4` and `dns_timeouts` operations in udns. This blocks the above bug at step 4.

I have very little experience with autoconf, make, etc. so the current build changes (which enable `HAVE_UDNS` always) are just a stub. They're a separate commit so they can be rebased out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ scripts/lt*.m4
 .#*
 \#*#
 *~
+*.swp
 
 # Packages #
 ############
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
       apt:
         packages:
         - libcppunit-dev
+        - libudns-dev
   - compiler: clang
     env: COMPILER=clang++-3.6 SKIP_COVERAGE=true
     addons:
@@ -20,6 +21,7 @@ matrix:
         packages:
         - clang-3.6
         - libcppunit-dev
+        - libudns-dev
   - compiler: clang
     env: COMPILER=clang++-3.7 SKIP_COVERAGE=true
     addons:
@@ -30,6 +32,7 @@ matrix:
         packages:
         - clang-3.7
         - libcppunit-dev
+        - libudns-dev
   - compiler: clang
     env: COMPILER=clang++-3.8 SKIP_COVERAGE=true
     addons:
@@ -40,6 +43,7 @@ matrix:
         packages:
         - clang-3.8
         - libcppunit-dev
+        - libudns-dev
   - compiler: gcc
     env: COMPILER=g++-4.7 SKIP_CHECK=true SKIP_COVERAGE=true
     addons:
@@ -47,6 +51,7 @@ matrix:
         sources: ubuntu-toolchain-r-test
         packages:
         - g++-4.7
+        - libudns-dev
   - compiler: gcc
     env: COMPILER=g++-4.7 SKIP_COVERAGE=true
     addons:
@@ -55,6 +60,7 @@ matrix:
         packages:
         - g++-4.7
         - libcppunit-dev
+        - libudns-dev
   - compiler: gcc
     env: COMPILER=g++-4.8
     addons:
@@ -63,6 +69,7 @@ matrix:
         packages:
         - g++-4.8
         - libcppunit-dev
+        - libudns-dev
 
 before_install:
   - if [ ! $SKIP_COVERAGE ]; then pip install --user cpp-coveralls; fi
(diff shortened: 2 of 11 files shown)


rakshasa:
I'll look at it later, however two points; if you make a new PR, close the old one. And also make sure that the Travis CI linked above is green.

rakshasa:
It might be an idea to add both with and without udns to the travis matrix, e.g. have all of them include udns and add two non-udns versions for 'env: COMPILER=clang++ SKIP_COVERAGE=true' and 'env: COMPILER=g++-4.7 SKIP_COVERAGE=true'.

slingamn:
There are a couple problems:

1. I realized that udns isn't on Travis's package whitelist. I've filed a request for them to add it: https://github.com/travis-ci/apt-package-whitelist/issues/3499
1. Right now, I don't have any automated support for enabling or disabling udns at compile-time, I just have `HAVE_UDNS` hardcoded in a header file. I was hoping for some help figuring out what to do here. I think ideally the build process would somehow encourage the user to build with `udns`, e.g., by requiring it unless an option to disable it is passed to `configure`?

In the meantime, I'm pushing a bug fix (invalid tracker hostnames could stop the client with `internal_error`) :-\

slingamn:
Seems like I'm waiting for some kind of cache invalidation re. the Travis whitelist.

slingamn:
Closing in favor of #134

PR #134, opened 2016-11-10
async DNS resolution via udns, take 3

slingamn:
This is #133, but with actual autoconf changes, and a green Travis build (with the two additional configurations you suggested).

Right now, the build requires udns by default and fails if it's not found; to build without udns, you need `--disable-udns`.


rakshasa:
When udns is not available it should not fail, it should compile as normal.

slingamn:
I implemented this with `AC_ARG_WITH` and `AC_CHECK_HEADERS`.

How would we convey to package maintainers that they should build in an environment with udns?

rakshasa:
If package maintainers want, or don't want, udns is up to them. That there is a check during configure that tests for udns by default is as far as libtorrent should go to force people to (not need to) make an active choice.

slingamn:
In my experience, without this patch and without the mitigation of a local caching DNS resolver, rtorrent is nearly unusable over a high-latency link (because of UI stalls) --- so I think it wouldn't hurt to recommend building with udns, or at least to document that building with udns is possible.

Did I address all the review feedback?

rakshasa:
It is basically a matter of different priorities.

While there are users who have issues with dns resolution blocking, there are also a lot of people compiling the client on weird setups where udns is not available, and even more where it is not installed by default.

Thus I do not want to make non-critical dependencies cause issues with compilation for other users. So if udns is installed, compile with it, if not then output a message during configure and compile normally.

slingamn:
I agree completely; I don't want to make it harder for end users to compile the project.

What I'm driving at is that I don't think most people are compiling the project themselves --- instead, they're getting it from the package manager of a distribution that also ships udns. So this is an opportunity to solve the problem in a centralized way that doesn't burden the individual user: if we document that it's recommended for package maintainers to build the project with udns support, then the maintainers can take care of ensuring that `udns.h` is available in the build environment and `libudns.so` is marked as a dependency of the `libtorrent` package. Then in these distributions, the package manager will take care of everything during install/upgrade, without burdening the user.

rakshasa:
I understand where you are coming from, and appreciate it.

However it is a decision that ultimately rests on me (unless there's a successful fork), and my decision remains in the favor of what I mentioned above. It is, after all, my inbox that is and gets flooded.

Focus instead on creating a good branch I can merge.

Which, btw, means fixing the incompatible indentation and such in e.g. 'udnsevent.cc', and moving that code to 'src/utils' as it is not a shared API. I'll follow that up with a more extensive review later.

slingamn:
I noticed that there is an issue with how these APIs (both the old synchronous API, and the new async one) handle single-stack settings. Suppose for example that the DNS server supports both A and AAAA records, but the client only has IPv4 connectivity. Then `getaddrinfo(3)` with `AF_UNSPEC` will return both IPv4 and IPv6 addresses, but `torrent::resolve_host` will arbitrarily choose the first one in line and pass it to the callback. Similarly, `UdnsEvent` will request A and AAAA records in parallel, but will call the callback with whichever is returned first. In each case, if the callback happens to be invoked with an IPv6 address, it will not succeed.

It seems like in order to fix this fully, we'd need a way to offer multiple addresses to callbacks like `UdpTracker::start_announce`, or else modify the state machine so that it first attempts to announce over IPv4 and then fails over to IPv6.

rakshasa:
I'm currently working on the 'feature-bind' branch that will ideally in the future be able to deal with those issues.

slingamn:
Anyway, I look forward to a more detailed review :-)

src/utils/udnsevent.cc
```diff
@@ -0,0 +1,215 @@
...
+    // TODO don't do this here --- do it once in the manager, or in rtorrent
+    ::dns_init(NULL, 0);
+    // thread-safe context isolated to this object:
+    m_ctx = ::dns_new(NULL);
+    int fd = ::dns_init(m_ctx, 1);
+    if (fd > 0) {
+      m_fileDesc = fd;
+    } else {
+      throw internal_error("dns_init failed");
+    }
```
rakshasa: Reorder this so that you only go into an if block in the exceptional case:

```
m_fileDesc = ::dns_init...;

if (fd == -1)
  throw ...
```

Also looking at the documentation it looks like if you call dns_init(m_ctx, 1) you don't need to call it above. Or if you do called it above, just use dns_open.

src/utils/udnsevent.cc
```diff
@@ -0,0 +1,215 @@
...
+
+    m_taskTimeout.slot() = std::bind(&UdnsEvent::process_timeouts, this);
+  }
+
+  UdnsEvent::~UdnsEvent() {
+    priority_queue_erase(&taskScheduler, &m_taskTimeout);
+    ::dns_close(m_ctx);
+    ::dns_free(m_ctx);
+
```
rakshasa: Set m_fileDesc to -1 here, it creates clear errors if calls are made to this object after it has been deleted.

src/utils/udnsevent.cc
```diff
@@ -0,0 +1,215 @@
...
+    int timeout = ::dns_timeouts(m_ctx, -1, 0);
+    if (timeout != -1) {
+      manager->poll()->insert_read(this);
+      manager->poll()->insert_error(this);
+      priority_queue_erase(&taskScheduler, &m_taskTimeout);
+      priority_queue_insert(&taskScheduler, &m_taskTimeout, (cachedTime + rak::timer::from_seconds(timeout)).round_seconds());
+    } else {
+      // no pending queries
+      manager->poll()->remove_read(this);
+      manager->poll()->remove_error(this);
```
rakshasa: Reorder this so you have:

```
if (timeout == -1) {
  ...
  return;
}

...
```

src/utils/udnsevent.cc
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+
+  void UdnsEvent::cancel(struct UdnsQuery *query) {
+    if (query == NULL) {
+      return;
+    }
```
rakshasa: Don't use brackets when the if block is a single line.

src/torrent/connection_manager.cc
```diff
@@ -48,11 +48,69 @@
...
+#else
+    MockResolve *mock_resolve = static_cast<MockResolve*>(query);
+    auto it = std::find(std::begin(m_mock_resolve_queue), std::end(m_mock_resolve_queue), mock_resolve);
+    if (it != std::end(m_mock_resolve_queue)) {
+      m_mock_resolve_queue.erase(it);
+      delete mock_resolve;
+    }
+#endif
+  }
+};
```
rakshasa: This seems a bit too much to include in this file, consider moving it to a source file in the same place as the udns stuff is.

slingamn:
This should address most of the issues. The second commit should be more readable since all the de-indentation noise is factored out.

There are two things I didn't do:

1. Moving `AsyncResolver` out of `connection_manager.cc`. It's just a proxy class that holds the implementation of some `ConnectionManager` methods (e.g., `ConnectionManager::enqueue_async_resolve`), so I thought moving it out would sacrifice locality.
1. Changing `m_malformed_queries` to be a `vector<unique_ptr>`. I thought this would confuse the object lifecycle of `udns_query`, because on some paths it has to be managed in a C-style way (in particular, when it does get passed to udns).

rakshasa:
Locality doesn't really work like that with current compilers, neither in terms of inline/static optimizations nor will library symbols be created as they are not declared as such.

Who is responsible for managing the lifecycle of malformed queries? From the code it seems it is the user, not the library.

slingamn:
Sorry, I was misusing the term "locality": I meant it in a code organization sense, not in the performance sense.

Anyway, `unique_ptr` makes sense for the mock resolver code in `AsyncResolver`, so I'm using it there now.

The issue is that the same `udns_query` object is used to store both valid and malformed queries, so memory management for both should follow the same patterns. My interpretation is that the library is managing the lifecycle: the structure is allocated by `enqueue_resolve` and then freed either automatically by callback execution or manually by `cancel`.

slingamn:
The build failed because LLVM's apt repository went down briefly. The rebuild passed: https://travis-ci.org/slingamn/libtorrent/builds/177980242

slingamn:
Need any follow-up from me on this?

rakshasa:
Yeah, sorry for the delay. The only real issue remaining is the code in connection_manager.cc and the use of ifdef's.

I'd prefer it if you made AsyncResolver into a base class with the sync and udns resolvers inheriting from it. Add static member functions that call the virtual versions and the only ifdef that should be needed in connection_manager.cc should be whetever to use sync or udns.

slingamn:
This isn't quite what you asked for, so feel free to tell me to change it. The additional level of indirection (public methods of `ConnectionManager` being thin wrappers around the public methods of `AsyncResolver`) was starting to seem redundant, so I exposed the `AsyncResolver` member directly as the interface. It's now a base class with virtual methods, although the subclass implementations still need to be wrapped overall in an `#ifdef` (since the `UdnsEvent` object can't actually be constructed without `udns.h`).

rakshasa:
I'll handle this once I'm done with the feature-bind branch.

slingamn:
Cool. If I understand correctly, this branch will need to be adapted so that `UdnsEvent` can resolve a name to multiple addresses, which will then be passed in to `bind_manager`?

rakshasa:
Yes, it seems we'll have to pass all the dns hits for an address when attempting a connection.

monosoul:
@rakshasa @slingamn 
Hello, guys! Any progress on this one?

rakshasa:
I'm still working on the feature-bind branch.

It is slow but steady progress, though due to the complexity of the feature it will take a while still to finish.

wolfen351:
just wanted to know if this is still happening?

slingamn:
I'm optimistic about this getting integrated once the feature-bind branch is complete, although it may take me some time to get warmed up again with the codebase.

The branch in this PR rebases cleanly on top of c167c5a9e0bcf (the current master) and works with the current rtorrent master; I'd encourage you to try it out if this is a change you're anticipating. Note that you must install `libudns-dev` or your distribution's equivalent before building libtorrent, otherwise udns support will be silently compiled out. Here's the recipe I've been using to build development versions of libtorrent/rtorrent:

````bash
mkdir -p ~/workspace/rtorrent_sandbox
cd ~/workspace/libtorrent
./autogen.sh
./configure --prefix ~/workspace/rtorrent_sandbox
make
make install

cd ~/workspace/rtorrent
export PKG_CONFIG_PATH=~/workspace/rtorrent_sandbox/lib/pkgconfig
./autogen.sh
./configure --prefix ~/workspace/rtorrent_sandbox
make
make install
````

This should suffice to build a development version of libtorrent (without disrupting your regular install), then build a development version of rtorrent that dynamically links against it. The resulting binary should be `~/workspace/rtorrent_sandbox/bin/rtorrent`.

**Comment by rakshasa** [comment]
@slingamn just wanted you to know that I've already pulled your branch and I'm working on it. I'll merge it soon.

**Comment by Kcchouette** [comment]
@rakshasa any news on the merge?
Thanks for your work :)

**Comment by rakshasa** [comment]
> @rakshasa any news on the merge?

Still working on the rtorrent-docker test environment, it's probably going to be a couple weeks still before it is ready.

**Comment by slingamn** [comment]
Just posting a quick merge/resolution with master, in case anyone else wants to use this patchset.

**Comment by Fossil01** [comment]
Any news on the merge?

**Comment by slingamn** [comment]
cc @stickz re. 25100693ee7ace605fbdbf58339b55af0d59b9f2

**Comment by arbitrary-dev** [comment]
+1 for waiting

**Comment by slingamn** [comment]
I verified that this applies against `v0.14.0` and runs stably, although it is unstable against a subsequent change (see #256 for details).

File: src/tracker/tracker_udp.cc [hunk 6]
```diff
@@ -70,25 +70,22 @@ TrackerUdp::TrackerUdp(TrackerList* parent, const std::string& url, int flags) :
 
   m_port(0),
 
-  m_slot_resolver(NULL),
   m_readBuffer(NULL),
   m_writeBuffer(NULL) {
 
   m_taskTimeout.slot() = std::bind(&TrackerUdp::receive_timeout, this);
+
+  m_resolver_callback = std::bind(&TrackerUdp::start_announce, this, std::placeholders::_1, std::placeholders::_2);
```
stickz, line 78: This is not stable. We can't have the callback in the `TrackerUdp` object. If the torrent is removed before the announce completes, it will trigger a crash. The callback will no longer be valid.

slingamn: Just to be clear, this line hasn't changed in years. You've been deploying this and it's been running stably, correct?

slingamn: Also if I understand your concern correctly, it's addressed by having `close_directly` remove the callback:

https://github.com/rakshasa/libtorrent/blob/a03859ee9fa9f65beeaec01c7694f8c8959568fa/src/tracker/tracker_udp.cc#L180-L182

stickz: Nope, we passing are `m_resolver_callback` as a pointer to enqueue. This must remain valid indefinitely because it's async. It was stable on my test environment. But once we tested it with 10k of clients, it started flooding crash dumps.
```c
  m_resolver_query = manager->connection_manager()->async_resolver().enqueue(
      hostname.data(),
      AF_INET,
      &m_resolver_callback
  );
```

This is the solution I implemented to avoid sending a pointer that could potentially become undefined. 
https://github.com/stickz/rtorrent/compare/32e1e6f6357d35b53aea947da16d9526fc5a3895...5517bef3cc71f1ac6191b89ae1194080385ddb18

slingamn: The callbacks are "asynchronous" in the sense that the main event loop does not block while the query is in flight, but the callbacks can only be triggered from `::dns_ioevent` or `::dns_timeouts`, which are called from within the main event loop (once the poll layer detects that a response is ready, or that the query has failed). Accordingly a call to `::dns_cancel` from the main event loop can prevent the callbacks from triggering. Because the main loop runs within a single thread, it should not be possible for these operations to race against each other.

According to my understanding, the specific scenario you described (DNS query is initiated, the torrent is removed, DNS response is received, the callback executes against a freed `TrackerUdp` object) cannot occur for this exact reason: `TrackerUdp::~TrackerUdp` calls `UdnsAsyncResolver::cancel`, which calls `::dns_cancel` on the callbacks pointing to the `TrackerUdp` object that is being freed.

I respect the intensive testing you have done on this patchset but I have run it continuously for years on heavy workloads without issue. I have also tested it with Valgrind, which found memory safety issues in the DHT layer (e.g. #166) but not with the udns code. I think it's possible that the issues you observed are a bad interaction with another patchset you're using? I would also be interested to see any backtraces or stack dumps you were able to collect.

stickz: Here's a stack trace I was able to collect and resolve for this problem.

```
Thread 1 (Thread 0x7f04816ad680 (LWP 114901)):
#0  0x00007f04820e350b in pthread_kill () from /usr/lib64/libc.so.6
#1  0x00007f0482086906 in raise () from /usr/lib64/libc.so.6
#2  0x00007f04820698b7 in abort () from /usr/lib64/libc.so.6
#3  0x000055de99b537c8 in do_panic (signum=11) at /usr/src/debug/net-p2p/rtorrent-0.9.8.5.4/rtorrent-5.4-0.9.8-0.13.8/rtorrent/src/main.cc:645
#4  <signal handler called>
#5  0x000055de9dd38e40 in ?? ()
#6  0x00007f0482613280 in std::function<void (sockaddr const*, int)>::operator()(sockaddr const*, int) const (__args#1=<optimized out>, __args#0=<optimized out>, this=<optimized out>, this=<optimized out>, __args#0=<optimized out>, __args#1=<optimized out>) at /usr/lib/gcc/x86_64-pc-linux-gnu/13/include/g++-v13/bits/std_function.h:591
#7  torrent::a4_callback_wrapper (result=<optimized out>, data=0x7f04541fc480, ctx=<optimized out>) at utils/udnsevent.cc:60
#8  torrent::a4_callback_wrapper (ctx=<optimized out>, result=<optimized out>, data=0x7f04541fc480) at utils/udnsevent.cc:40
#9  0x00007f0481addf1c in dns_ioevent () from /usr/lib64/libudns.so.0
#10 0x00007f04825a9e9e in torrent::PollEPoll::perform (this=0x55de9ac6b630) at torrent/poll_epoll.cc:185
#11 0x00007f04825d41b4 in torrent::thread_base::event_loop (thread=0x55de9ac72300) at torrent/utils/thread_base.cc:174
#12 0x000055de99b7aa5c in main (argc=<optimized out>, argv=<optimized out>) at /usr/src/debug/net-p2p/rtorrent-0.9.8.5.4/rtorrent-5.4-0.9.8-0.13.8/rtorrent/src/main.cc:508
```

slingamn: Thanks for getting back to me on this.

I believe this is a bug in your fork. You originally merged the udns changes in https://github.com/stickz/rtorrent/commit/62d3f5e2a81a666f8b708a740142339b0557ff04 ; this version of the patch correctly includes the call to `AsyncResolver::cancel` in `TrackerUdp::close_directly` that I linked above. Then subsequently in https://github.com/stickz/rtorrent/commit/a843234b54372fd89166ad75a737a4e5f79f4284 you replaced `tracker_udp.cc` with a version (from libtorrent's master?) that does not include this call. Consequently, the callback does not get cancelled and can fire even when it has become invalid, which is what we see in the stack trace.

slingamn:
@rakshasa what would it take to get this merged? Is `feature-bind` still in progress?

The current behavior in this branch is to do IPv4 resolution only.

rakshasa:
There's too many easier things to fix atm, and I'd like to do a more comprehensive review of the current tracker code first.

This includes fixing some ipv4/6 issues I'm seeing in rtorrent-docker.

rakshasa:
I've started work on making the tracker code thread-safe.

This will allow tracker request to be done in a separate thread, and once that is done I'll implement async dns resolution using std::async in a dns-specific thread.

slingamn:
It's up to you as the maintainer but I think the approach here has some strong advantages that offset the use of an external dependency. It seems like there are two possibilities for a threaded implementation:

1. One thread per DNS lookup. This has obvious scalability problems for deployments with thousands of torrents.
2. One thread that does all the DNS lookups, via synchronous `getaddrinfo(3)` and message passing in a loop. This also has scalability problems for high-latency links (e.g. if each lookup takes 20 milliseconds even in the best case).

udns was designed specifically for easy integration with applications like rtorrent (it exposes a single FD that can be monitored from the same event loop as everything else), and the implementation here has been performant and stable for eight years now.

rakshasa:
I wouldn't create a new async thread for each request, there would be worker threads started by async as necessary, and they would be limited.

However udns is LGPL, so the source code can be directly included in libtorrent which is GPL.

So that's an option.

rakshasa:
#334

Reimplemented udns support, no longer needs to install the library as it is included with libtorrent and the code is much cleaner.

PR #143, opened 2016-12-21
Fix the build with OpenSSL 1.1.

ppentchev:
Hi,

What do you think about the attached patch that fixes the build with OpenSSL 1.1, where the DH structure is now opaque and we cannot set its member fields directly?

Thanks for taking care of libtorrent and rtorrent, and keep up the great work!

G'luck,
Peter


rakshasa:
A modified version of this patch has already been applied to the feature-bind branch.

PR #148, opened 2017-03-09
uniform_rng - Uniform distribution random number generator

pyroscope:
This is the basis for a `system.rand` rtorrent command, and also a pre-cursor to fix entropy problems the numerous `random() % limit` calls have – namely using the lower low-quality bits of the PRNG, and having potential (or real) thread-safety problems.

:bangbang: When testing the original `system.rand`  implementation based on `random()`, the number sequences were awfully unrandom.

### DIFF
--- a/src/utils/uniform_rng.cc
+++ b/src/utils/uniform_rng.cc
@@ -1,0 +1,93 @@
+// libTorrent - BitTorrent library
+// Copyright (C) 2005-2017, Jari Sundell
+//
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; either version 2 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, write to the Free Software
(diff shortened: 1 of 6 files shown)


src/utils/uniform_rng.cc
```diff
@@ -0,0 +1,93 @@
...
+
+#include "config.h"
+#include "globals.h"
+#include "torrent/exceptions.h"
+
+
+namespace torrent {
+
+uniform_rng::uniform_rng() {
+    unsigned int seed = cachedTime.usec() ^ (getpid() << 16) ^ getppid();
```
ss23: Could we seed with `RAND_bytes` if rtorrent was compiled with OpenSSL support? I realise this isn't meant to be cryptographically secure, but I don't think it'll hurt to make sure we're using a high quality seed.

src/utils/uniform_rng.h
```diff
@@ -0,0 +1,72 @@
...
+class uniform_rng {
+public:
+    uniform_rng();
+
+    int rand();
+    int rand_range(int lo, int hi);
+    int rand_below(int limit) { return this->rand_range(0, limit-1); }
+
+private:
+    char m_state[128];
```
ss23: Given people might have their client running for a very long time, could we change this to 256 bits to help with potential issues?

ss23:
Was the plan to make another PR after this using the new code where random() is currently used?

rakshasa:
I'm going to move to c++11 only, so use the random number generators it provides.

rakshasa:
Going to close this as random number generation is going to be handled by "torrent/utils/random.h" in feature-bind.

It should be using thread local global/static generator however I've not gotten around to implementing it yet.

PR #149, opened 2017-03-26
[RFC] Use native alignment and the standard allocator

anthonyryan1:
This pull request is primarily to discuss the benefits and drawbacks of manually managing the alignment of memory rather than allowing the compiler and libc to do it themselves.

While this code seems to have originally been written as a performance improvement. In my testing I haven't actually seen it do a better job than simply allowing the compiler to do it's job. Furthermore I observed a repeatable memory usage reduction of just under 4% when averaging approximately 100 instance runs downloading a file each time. I did not see a significant change in CPU usage.

As always, I look forward to your thoughts.

-------

The cacheline allocator and alignment hints passed to the compiler
rely upon /usr/include/linux/cache.h, a file which is no longer
present on any linux distro I checked. This resulted in it using
the safe fallback of 128 byte alignment.

Unfortunately, 128 byte alignment is oversized for a wide range of
structures and types we're enforcing this alginment on. Furthermore
we're occasionally passing these properties to functions that are not
expecting the 128 byte alignment.

UBSan in Clang 3.7 found ~230 sections where 128 byte aligned objects
were handled by code that wasn't expecting them to be 128 byte aligned.
This doesn't appear to cause issues in practice because the generous
padding is simply ignored.

The same issue exists within rtorrent as well, and a separate pull
request for it will be created at a later date if this is approved.


rakshasa:
Cacheline is used in libtorrent to ensure atomic updates of data and structures without needing the use of mutex and other expensive locking mechanisms.

PR #152, opened 2017-05-15
Fix honoring system.file.allocate.set=1 rtorrent config setting - clean base

chros73:
For more info: https://github.com/rakshasa/rtorrent/pull/604

Depended by: https://github.com/rakshasa/rtorrent/pull/604
Probably clashes with: #147 (it depends on it)

Refers to: https://github.com/rakshasa/libtorrent/pull/109


trim21:
I think we can close this too

PR #164, opened 2018-02-05
Add support for webseeds

naf419:
Initial attempt.

I tried to follow the psuedo-spec at http://getright.com/seedtorrent.html

(Barely) Tested against archive.org samples.


chros73:
Looks interesting! Let me know when you feel it'ts ready to be tested and I'll include it in rtorrent-ps-ch on a branch to be able to test it and I'll post on the original issue.

@rakshasa , this one needs your guidance. :)

@naf419, Thank you for your work!

naf419:
@chros73, well before you try it, let me tell you what i think it does:

it should:
- throw a dummy tracker on the tracker list for each url in the url-list
- download 1 needed torrent piece at a time from a randomly selected url from the url-list via whatever-protocol-curl-supports (only tested against http)
- remove urls from the url-list the first time it receives an error (cant connect / not found / bad hash)
- stop trying when it has no un-errored urls left
- add to the torrent's and the global download speed indicators
- cleanly finish current piece and stop when torrent is stopped

it wont:
- honor the max download choke speed
- handle threads eloquently. currently spawns 1 thread per torrent-with-webseeds.

given that, if you'd like to test it, have at it.

chros73:
Thanks for the detailed comment. Do youbplan to modify the functionalty further or this is the last take?

naf419:
@chros73 no planned functionality changes

chros73:
I couldn't compile libtorrent :( : 
- your patch is fine (luckily without any conflicts)
- but the building env (set up by the build script) is not: libtorrent doesn't find the (headers of) rpath-linked libcurl during compilation
  - the weird thing is that rtorrent does (it also uses the same libcurl)

I couldn't find the cause of the issue "quickly". I don't/won't have time to "play" with it in the near future, sorry.
If I managed to figure out something I'll let you know.
Thanks for your work!

chros73:
I played with it today and managed to compile it:
- I've modified [configure.ac](https://github.com/chros73/rtorrent-ps-ch/commit/8390a126594fbecde59e8bfc314eff0f3aaab6da#diff-7801f81dc7c47445ee3dc358586b1a7bR1) of your patch
- I've also added 1 more line into [libtorrent.pc.in](https://github.com/chros73/rtorrent-ps-ch/commit/8390a126594fbecde59e8bfc314eff0f3aaab6da#diff-7801f81dc7c47445ee3dc358586b1a7bR42)

I had to add back setting [CPPFLAGS](https://github.com/chros73/rtorrent-ps-ch/commit/17a3c90fd126c80e4356cc0f324916df1a32fe26#diff-0b83f9dedf40d7356e5ca147a077acb4R155) into build script, otherwise the include CFLAGS doesn't get passed to libtool for some reason that I still don't understand why:
- the generated config.status, config.log, libtorrent.pc, etc seems fine and still libtool doesn't use the specified include CFLAGS directive during compilation otherwise

chros73:
So, I tried out with an archive.org torrent and it seems to be working, but when I tried to create my own (creating with utorrent 2.2.1) it didn't download it. Have you tried it with your own torrent file?

And does it deal with multi/single-file torrents as well, e.g. as in [transmission](https://github.com/transmission/transmission/blob/master/libtransmission/webseed.c#L444)?

> it won't honor the max download choke speed

Probably you meant throttle groups, and not choke groups.
It'd be nice if:
- it would add 1 peer / working url to the peerlist

naf419:
@chros73 i made some torrents myself and you are right about it not working. i only handled when multiple urls were present. and i didn't handle url's not ending with a / either even then.
fix coming shortly.

naf419:
@chros73 according to my reading of the getright spec, url-list entries can either be:
`http://srv/folder/filename`  <-- no trailing /
or
`http://srv/folder/` <-- trailing /, so torrent name appended to url

multi-file torrents files follow the same rule, but also add the sub-file name to the end of the url.

your TP-LINK-TL-WR842N-ND-V2.torrent didnt work because it ends in a slash, but already has the torrent name included in the url prefix.

chros73:
Errr, that webseed url doesn't have a trailing slash.... Or am I mistaken?

naf419:
Ya, the one that didn't work for you had a trailing slash. Should have been like this (the url you wrote without the /) [TP-LINK-TL-WR842N-ND-V2-corrected.zip](https://github.com/rakshasa/libtorrent/files/1738230/TP-LINK-TL-WR842N-ND-V2-corrected.zip)

chros73:
You were right, sorry (I haven't looked in the generated torrent file): utorrent 2.2.1 added the trailing slash. :(
It works indeed fine after editing the file manually. (I also updated my above comments.)

So, it seems all the URL related issues are good now.
What happens if there are other seeders in the swarm? Will webseed be used?

A. I suggest to modify `configure.ac`:
- PKG_CONFIG_MODULES related [lines](https://github.com/chros73/rtorrent-ps-ch/blob/feature/webseed/patches/backport_0.13.7_04-webseed_support.patch#L15) : 
  - curl is required now
- and also add [this](https://github.com/chros73/rtorrent-ps-ch/blob/feature/webseed/patches/backport_0.13.7_04-webseed_support.patch#L7) , makes more checking

B. Although a lot still has to be done if you (we) want this to get merged. :(

May I suggest to go through the web.* and webseed.* files of [transmission](https://github.com/transmission/transmission/tree/master/libtransmission) and "steal" ideas to use in your tracker/controller, e.g. like [this](https://github.com/transmission/transmission/blob/master/libtransmission/webseed.c#L64)?

I think these classes have to be finished first and then the rest of the integration can be dealt with.

Thanks for your work!

naf419:
> What happens if there are other seeders in the swarm? Will webseed be used?
Yes, it gets the next desired chunk from the same pool of desired chunks as the peer requests do, so it should [untested] work fine in simultaneous cooperation with other seeding peers.

RE: merge, i don't imagine it ever will, I just threw it up here for anyone who wanted to patch it in themselves

rakshasa:
Closing this as it's not going to be merged.

There's going to be a lot of work that needs to be done before webseeding can be supported.

I'm willing to prioritize this if supported by donations, however it is not a priority for me atm as it there are low-hanging fruits to be picked.

PR #189, opened 2019-06-04
"Warning message" from some trackers when a torrent isn't registered.

montoner0:
Some trackers send "warning message" instead of a failure when a torrent is not registered on the tracker. Looks like this: 

> d8:intervali3030e12:min intervali3030e5:peers6:S‹™жЊЄ15:warning message22:Torrent not registerede

The change will turn this warning into an error.

### DIFF
--- a/src/tracker/tracker_http.cc
+++ b/src/tracker/tracker_http.cc
@@ -272,6 +272,17 @@ TrackerHttp::receive_done() {
 
   // If no failures, set intervals to defaults prior to processing
 
+  if (b.has_key("warning message") && b.get_key("warning message").is_string()){
+
+    std::string msg=b.get_key_string("warning message");
+
+    if ( (msg.find(" not registered")!=std::string::npos) ||
+    		(msg.find("unregistered")!=std::string::npos) ||
+     		(msg.find("torrent cannot be found")!=std::string::npos) )
+
+       return receive_failed("Warning message \"" + msg + "\"");
+  }
+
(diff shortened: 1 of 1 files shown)


rakshasa:
> warning message: (new, optional) Similar to failure reason, but the response still gets processed normally. The warning message is shown just like an error.

I don't think it would be correct to fail the tracker reply based on words in the warning message, it should instead have a way of both processing the reply and displaying the message in the client.

montoner0:
Agree, it would be nice to have some more flexible solution. But for know, as a quick hack it should be enough. Or it would be better to register this as an issue?

kannibalox:
I'm guessing the one of the reasons tracker admins were even using `warning message` instead of `failure message` was to work around rTorrent's behavior of ignoring intervals on properly-bencoded tracker failures. Merging this would mean it would go back to the "announce every 300 seconds" pattern, instead of the `3030` being requested in the response you posted. I've opened a PR here to tell libtorrent to respect `min interval` here https://github.com/rakshasa/libtorrent/pull/246, but even with that fix, I'm not sure it's worth playing whack-a-mole in terms of which keys are error codes and which aren't.

rakshasa:
To implement a proper solution for this, can someone post examples of warning messages from e.g. `tracker_dump`.

I can add (very strict) handling of those and have them e.g. set a flag you can check with xmlrpc commands and have it show it in the CLI client.

rakshasa:
Added.

PR #190, opened 2019-07-07
checks.m4: Fix fallocate check

neheb:
fallocate needs _GNU_SOURCE defined


chros73:
On which distribution?

rakshasa:
Fixed.

PR #194, opened 2019-07-21
Get rid of NULL

neheb:
nullptr is the C++11 replacement for NULL. Found with
-Wzero-as-null-pointer-constant .

Also changed a few C headers to use the C++ versions as the C headers
cause the same warning to be emitted.


rakshasa:
Two things; first, LLVM has an amazing tool to update code to c++11++ automatically.

Second, doing this will break all patches so not happening now. I'll be upgrading the code when more stuff has been merged.

PR #195, opened 2019-07-25
Allow log files to be appended

Nicholi:
Rather than current implentation of truncate/overwrite by default.

The only real difference from standard open_file is in the ofstream constructor. Added tests as well.

Makes using simple file names great again, and prevent rtorrent from overwriting all files on a restart. I also ran into a small issue when using logrotate.d and the copytruncate directive. After a log rotation had been performed, and when rtorrent began writing to file again, it wouuld be filled with null 0's up to the last position it had been written to. This would potentially solve that.

### DIFF
--- a/src/torrent/utils/log.cc
+++ b/src/torrent/utils/log.cc
@@ -73,7 +73,7 @@ struct log_cache_entry {
 };
 
 struct log_gz_output {
-  log_gz_output(const char* filename) { gz_file = gzopen(filename, "w"); }
+  log_gz_output(const char* filename, bool append) { gz_file = gzopen(filename, append ? "a" : "w"); }
   ~log_gz_output() { if (gz_file != NULL) gzclose(gz_file); }
 
   bool is_valid() { return gz_file != Z_NULL; }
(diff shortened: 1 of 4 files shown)


rakshasa:
If we're adding append it should also be done for gzip'ed logs.

Instead of a separate function, pass append as a parameter to both log_open_file and log_open_gz_file.

Nicholi:
Not the greatest C dev, but I'm guessing that should just mean passing 'a' to gzopen?
The concept of appending to a gzip seems a little weird, but if it works.

I'll re-work the PR to take that into account, both as a parameter. And change up my PR related to this on rtorrent as well.

chros73:
@Nicholi , what do you think about @pyroscope's [solution](https://github.com/pyroscope/rtorrent-ps/blob/master/patches/lt-ps-log_open_file-reopen_all.patch) instead?
It seems way easier, not requireing any new command. I'm not sure about gz type though.

rakshasa:
@chros73 That has already been merged, this is a different issue as people might want to keep the previous log file content.

Also yes, just add a bool parameter and "a" to gzopen should work.

chros73:
@rakshasa , Oh, I see, I can't keep track anymore what's happened :)

Nicholi:
Switched to simple append parameter, with default value of false. Should prevent any API breakage as well.

PR #201, opened 2019-08-23
Replace usleep with C++11's sleep_for

neheb:
usleep is deprecated and replaced by nanosleep in POSIX 2008.

sleep_for is standard C++11 and internally uses nanosleep.

### DIFF
--- a/src/torrent/utils/thread_base.cc
+++ b/src/torrent/utils/thread_base.cc
@@ -1,8 +1,9 @@
 #include "config.h"
 
+#include <chrono>
 #include <cstring>
 #include <signal.h>
-#include <unistd.h>
+#include <thread>
 
 #include "torrent/exceptions.h"
 #include "torrent/poll.h"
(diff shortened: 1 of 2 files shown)


neheb:
Build failure seems unrelated.

neheb:
Rebased.

rakshasa:
Add this as a member function to `thread_base` and also update unit tests.

neheb:
On further evaluation, these should use condition variables, no?

src/torrent/utils/thread_base.cc
```diff
@@ -60,7 +61,7 @@ thread_base::stop_thread_wait() {
   release_global_lock();
 
   while (!is_inactive()) {
-    usleep(1000);
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
uis246, line 64: Looks like average thread join. Sleeping for join is anti-pattern in userspace.

neheb: correct

PR #215, opened 2020-12-08
Fix file truncation on resume with fallocate enabled

jmakovicka:
- Call ftruncate() after fallocate() to ensure the file gets shrunk
  if needed.

- Open the files queued for resizing in writable mode to allow
  fixing the files with a wrong size later.

### DIFF
--- a/src/data/socket_file.cc
+++ b/src/data/socket_file.cc
@@ -107,14 +107,14 @@ SocketFile::set_size(uint64_t size, int flags) const {
     throw internal_error("SocketFile::set_size() called on a closed file");
 
 #ifdef HAVE_FALLOCATE
-  if (flags & flag_fallocate && fallocate(m_fd, 0, 0, size) == 0)
+  if (flags & flag_fallocate && fallocate(m_fd, 0, 0, size) == 0 && ftruncate(m_fd, size) == 0)
     return true;
 #endif
 
 #ifdef USE_POSIX_FALLOCATE
   if (flags & flag_fallocate &&
       flags & flag_fallocate_blocking &&
-      posix_fallocate(m_fd, 0, size) == 0)
+      posix_fallocate(m_fd, 0, size) == 0 && ftruncate(m_fd, size) == 0)
     return true;
 #endif
 


slingamn:
Did you observe a problem with this in practice? I am not sure that this is actually necessary. `man 2 fallocate` says:

>This default behavior closely resembles the behavior of the posix_fallocate(3) library function, and is intended as a method of optimally implementing that function.

`man 3 posix_fallocate` says:

>If the size of the file is less than offset+len, then the file is increased to this size; otherwise the file size is left unchanged.

This is the behavior I observed in practice from `fallocate(2)`. More specifically, `man 2 fallocate` says:

>Because allocation is done in block size chunks, fallocate() may allocate a larger range of disk space than was specified.

however, I understand this as referring to the length of the *allocation*, not the *file size*. (Compare the behavior of `fallocate(2)` with `FALLOC_FL_KEEP_SIZE` where it increases the size of the allocation without changing the file size. Both versions of the call will typically result in an allocation larger than the file size, but without `FALLOC_FL_KEEP_SIZE` the allocation is just rounded up to the nearest block.)

slingamn:
Thanks for the explanation. It seems like these changes are relevant even when `system.file.allocate` is not enabled in rtorrent (and consequently `flag_fallocate` is not set)?

src/torrent/data/file_list.cc
```diff
@@ -592,6 +592,7 @@ FileList::create_chunk_part(FileList::iterator itr, uint64_t offset, uint32_t le
   // Check that offset != length of file.
 
   if (!(*itr)->prepare(prot))
+  if (!(*itr)->prepare(prot | ((*itr)->is_resize_queued() ? MemoryChunk::prot_write : 0)))
```
rakshasa, line 595: Besides the duplicate if statement, adding `prot_write` is wrong.

open_file is always opened initially as read-only for a reason, it is not the stage where we verify or start changing files.

If truncation is required it needs to request write access explicitly at a higher level rather than depend on requesting write access on all calls to create chunk.

Reason why this was written as is was that we shouldn't truncate files that might have belonged to other downloads without explicit permission.

rakshasa:
See #215 for new PR.

Issue #219, opened 2021-05-26
libtorrent: FTBFS on hurd-i386: error: 'IPV6_TCLASS' was not declared in this scope.


hahawang-github:
I decide to fix a broken package found at the recommended
page(https://people.debian.org/~sthibault/out_of_date.txt) named `libtorrent`.

After I downloaded the package source and tried to build without any modifications under the debian hurd running in qemu (debian-hurd-20210219.img), I got the following error.

```
   socket_fd.cc: In member function 'bool torrent::SocketFd::set_priority(torrent::SocketFd::priority_type)':
   socket_fd.cc:78:43: error: 'IPV6_TCLASS' was not declared in this scope; did you mean 'IPOPT_CLASS'?
```

To fix this bug, thanks to the help of people at debian-hurd & bug-hurd mailing list,  I decide to block the usage of the `IPV6_TCLASS` when it is not defined.

A pull request is available.

The gnu hurd mailing list, that already contains a submitted patch and other people's reviews.

https://lists.gnu.org/archive/html/bug-hurd/2021-04/msg00080.html

rakshasa:
Closing, see PR.

PR #220, opened 2021-05-26
Fix bug:src/net/socket_fd.cc:disable the usage of IPV6_TCLASS if it i…

hahawang-github:
I have already report this bug to the debian bug tracking system, debian-hurd and bug-hurd.
With their advises, I decide to block the use of `IPV6_TCLASS` under the GNU Hurd.

https://github.com/rakshasa/libtorrent/issues/219


rakshasa:
This seems like too obscure of a bug to be merged as a fix that adds complexity to the libtorrent codebase.

PR #221, opened 2021-06-21
Feature/detect ip

rakshasa:


anthonyryan1:
@rakshasa I noticed this removes `&ipv4=` and `&ipv6=` from announcements. I'm seeing behaviour regressions on dual stack clients where trackers are only receiving an IPv6 address now.

Looking over BEP7, it has been updated such that the new recommendation is to announce to the tracker once on IPv4 and once on IPv6. That unfortunately doubles the load burden on trackers.

Would you agree with that strategy and be open to a patch? Or how do you feel we should best proceed.

anthonyryan1:
One additional consideration, is that if we announce multiple times, private trackers will double-count upload and download.

Making that idea probably dead-on-arrival for torrents with the private flag.

PR #241, opened 2023-03-27
Improve the file README

WagnerGMD:
The principal purpose is to provide the instruction to build LIbTorrent because they are difficult to find for a lot of people !

From my point of view, you can close (or merge ?) t[he 230 PR](https://github.com/rakshasa/libtorrent/pull/230/) because it doesn't provide : 
- the complete instructions to install LIbTorrent (it's require the line `autoreconf -fi` but one full example, it's always better).
- neither the both example because the master and the latest release are different !
- usually the documentation are also include inside this file (specially true when we can't find it (an example : the wiki doesn't exist)).
- and for the record, the file autogen.sh didn't exist anymore (vanished a few years ago).

### DIFF
--- a/README
+++ b/README
@@ -1,27 +1,22 @@
-LibTorrent
-
-Copyright (C) 2005-2014, Jari Sundell
-
-LICENSE
-
- GNU GPL, see COPYING. "libtorrent/src/utils/sha_fast.{cc,h}" is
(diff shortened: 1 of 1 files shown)


rakshasa:
Closing this as it's not going to be merged.

PR #245, opened 2023-09-12
Ensure files starting on a new piece aren't marked complete twice

kannibalox:
Fixes https://github.com/rakshasa/rtorrent/issues/1248

### DIFF
--- a/src/torrent/data/file.h
+++ b/src/torrent/data/file.h
@@ -78,6 +78,7 @@ public:
   bool                has_permissions(int prot) const          { return !(prot & ~m_protection); }
 
   uint64_t            offset() const                           { return m_offset; }
+  uint64_t            offset_end() const                       { return m_offset + m_size; }
 
   uint64_t            size_bytes() const                       { return m_size; }
   uint32_t            size_chunks() const                      { return m_range.second - m_range.first; }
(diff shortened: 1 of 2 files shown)


stickz:
This pull request has a regression where the completion file list is reported as less than 100% when completed.

kannibalox:
Is there any further information you can provide? The size and order of the files in question and the piece size would be great. I tried a bunch of things, including setting some of the files to "off" priority, but was unable to reproduce.

kannibalox:
That should cover all cases, including zero-length files, but I'll do further testing and maybe even take a stab at unit tests to prove it out.

stickz:
Thanks. I'm working on pushing some of these fixes to various vendors. I backported changes from jesec as well.

stickz:
This pull request is still broken. Here's an example where the file completion percentage reports incorrectly.
https://github.com/swizzin/swizzin/issues/1106

Running the same code with a patch. 
https://github.com/swizzin/swizzin/blob/master/sources/patches/rtorrent/piece-boundary-fix-0.13.8.patch

kannibalox:
Very strange, I made sure to test what this condition at least looks to be and it functions correctly, I'll have to find a way to reproduce it.

kannibalox:
Closing for now, as this approach to the fix still has some weird edge cases that need working out.

PR #246, opened 2023-09-22
Respect min interval returned by the tracker for failed announces

kannibalox:
This changes the default values for the base class to `0`, and relies on the successful parsing of an announce to set them to something sane. If the interval can't be parsed during a failure (from an unrelated failure or just the keys missing), it fails back to the original backoff behavior. However, in order to accommodate the previous logic without modifying the ABI, the backoff timeout is capped at 299 seconds instead of 320.

Fixes [#1215](https://github.com/rakshasa/rtorrent/issues/1215)

### DIFF
--- a/src/torrent/tracker.cc
+++ b/src/torrent/tracker.cc
@@ -51,8 +51,8 @@ Tracker::Tracker(TrackerList* parent, const std::string& url, int flags) :
   m_group(0),
   m_url(url),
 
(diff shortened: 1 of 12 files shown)


Pentaphon:
@rakshasa please make this a high priority for private trackers

rakshasa:
Merged. #260

PR #247, opened 2023-10-17
Add basic padding file support

kannibalox:
This is a very simple partial implementation of [BEP 47](https://www.bittorrent.org/beps/bep_0047.html), adding simple padding file support (`p` in the `attr` key). It's not feature complete in that rTorrent may still unnecessarily request data from peers, but the padding files are:
* not stored on disk (instead represented by an anonymous mmap when reading/writing)
* allowed to exist as duplicate "files" (addresses https://github.com/rakshasa/rtorrent/issues/1156)


rakshasa:
Merged. #301

PR #248, opened 2023-11-29
Simplify updating queue entries

kannibalox:
The erase/insert pattern can be replaced by a single method that updates an existing entry's timer, or inserting if it doesn't exist.

This primary goal was to reduce CPU usage when handling incoming connections on large instances, but since the optimization was pretty generic it's been applied to many places.

### DIFF
--- a/rak/priority_queue.h
+++ b/rak/priority_queue.h
@@ -74,6 +74,10 @@ public:
     base_type::pop_back();
   }
 
+  void update() {
+    std::make_heap(begin(), end(), m_compare);
+  }
+
   void push(const value_type& value) {
     base_type::push_back(value);
     std::push_heap(begin(), end(), m_compare);
@@ -95,12 +99,8 @@ public:
     return true;
   }
 
-  // Removes 'itr' from the queue. This assumes 'itr' has been
-  // modified such that it has a higher priority than any other
-  // element in the queue.
+  // Removes 'itr' from the queue.
   void erase(iterator itr) {
-//     std::push_heap(begin(), ++itr, m_compare);
-//     pop();
     base_type::erase(itr);
     std::make_heap(begin(), end(), m_compare);
   }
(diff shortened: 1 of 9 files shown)


stickz:
Great Job @kannibalox. This pull request is 100% stable and greatly increases the number of peers rTorrent can process! It will allow over 10 gigabits to be pushed through rTorrent, provided the solid state hardware and system memory specs exist.

Is there any chance you could reduce CPU usage for downloading? I can't obtain any faster than 2 gigabits because it's CPU bound. I believe the problem is related to the way libtorrent rakshasa manages the pieces.

kannibalox:
> Is there any chance you could reduce CPU usage for downloading? I can't obtain any faster than 2 gigabits because it's CPU bound. I believe the problem is related to the way libtorrent rakshasa manages the pieces.

I did find a couple other things that could be optimized, but without a profile of actual usage from `perf record` or something similar, it's hard to know where the busiest parts are, and I didn't want to flood the repo with lots of micro-optimization PRs

stickz:
> > Is there any chance you could reduce CPU usage for downloading? I can't obtain any faster than 2 gigabits because it's CPU bound. I believe the problem is related to the way libtorrent rakshasa manages the pieces.
> 
> I did find a couple other things that could be optimized, but without a profile of actual usage from `perf record` or something similar, it's hard to know where the busiest parts are, and I didn't want to flood the repo with lots of micro-optimization PRs

@kannibalox Please send me an email. stickman002@mail.com
I have what you need. I don't want to post sensitive data here.
I recorded 100% CPU usage at 4.3ghz downloading at 2gbits.

stickz:
@kannibalox Thank you for your email. I have determined the download delegator which determines the pieces to queue is responsible for the high CPU usage. There are multiple `std::find_if` search patterns being called very rapidly at 2gbps.
https://github.com/rakshasa/libtorrent/blob/master/src/download/delegator.cc

anthonyryan1:
@rakshasa are you able to test and review this at your convenience?

stickz:
> @rakshasa are you able to test and review this at your convenience?

I've done extensive testing with this pull request, it should be ready to go.

rakshasa:
Merge with 7730b2561d381b004e6e29fa94b7f96b44e06b2a

PR #251, opened 2024-08-21
Remove decision structure for popcount_wrapper

stickz:
The decision structure is very costly. It's significantly faster to work with long long than to use it.


stickz:
@rakshasa Any thoughts about this PR? It's at least 15% faster in all circumstances. The decision structure is very costly due to the large number of times it's invoked. It's better just to work with the larger datatype.

rakshasa:
Merged, thanks.

PR #252, opened 2024-09-01
Increase max throttle speed to 33 gigabits

stickz:
This commit increases the max throttle speed from 1.6 gigabits to 33 gigabits.

It's no longer possible to use a bit shift operator and we must use some 64 bit unsigned integers.

This is critically useful for ruTorrent to restore throttle functionality, since many connections now exceed 1.6 gigabits.


rakshasa:
Merged.

PR #262, opened 2024-11-29
Remove unused code and convert some functions to lamdas/std::fun

kannibalox:
All the code in functional.h should be able to be replaced by the C++11 constructs. This first PR is more just demonstration, removing unused functions and replacing a couple of the lesser-used ones to ensure a) these patterns are something you're good with and b) before the changes need to happen across a ton of different files. If you're good with the changes, I'd be happy to put all them in this PR, or create multiple medium-sized ones, etc, whatever you'd prefer.

### DIFF
--- a/rak/functional.h
+++ b/rak/functional.h
@@ -52,21 +52,6 @@ struct reference_fix<Type&> {
   typedef Type type;
 };
 
-template <typename Type>
-struct value_t {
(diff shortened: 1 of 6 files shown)


rakshasa:
I've wanted to remove everything in the rak include folder for years, so this is a good start.

Issue #256, opened 2024-11-10
Caught internal_error: HandshakeManager::receive_succeeded(...) Unread data won't fit PCB's read buffer.


slingamn:
Congratulations on the new release!

I tried applying the patchset from #134 to master (at the time of writing, eddcb1f59105bd0ad2377b477d58641a1c6c9f67). (Without the patchset, my rtorrent instance is basically unresponsive due to the high number of synchronous DNS lookups). rtorrent then crashes immediately with:

```
Caught internal_error: HandshakeManager::receive_succeeded(...) Unread data won't fit PCB's read buffer.
/home/shivaram/workspace/sandbox/lib/libtorrent.so.22(_ZN7torrent14internal_error10initializeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x74) [0x74b6be624484]
/home/shivaram/workspace/sandbox/bin/rtorrent(_ZN7torrent14internal_errorC1EPKc+0x88) [0x593e64ff8af8]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.22(+0x460b4) [0x74b6be5eb0b4]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.22(+0x45988) [0x74b6be5ea988]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.22(_ZN7torrent9PollEPoll7performEv+0xe9) [0x74b6be62c5f9]
/home/shivaram/workspace/sandbox/lib/libtorrent.so.22(_ZN7torrent11thread_base10event_loopEPS0_+0x12a) [0x74b6be62014a]
/home/shivaram/workspace/sandbox/bin/rtorrent(+0x47c60) [0x593e64ff0c60]
/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca) [0x74b6bde2a1ca]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b) [0x74b6bde2a28b]
/home/shivaram/workspace/sandbox/bin/rtorrent(+0x485d5) [0x593e64ff15d5]
```

I bisected this to 11c38cc540bc1ca7c193cdc28be6e9890df4c84b / #254. So the release tag `v0.14.0` and the subsequent commit a83dcb9e667d65035084762b72279f4a43d1e565 don't have this problem. But it's not clear to me what in #254 would cause this.

Thanks for your time.

rakshasa:
[...]

PR #264, opened 2024-12-01
fix buffer position tracking in read_port

slingamn:
#254 / 11c38cc540bc1ca7c193cdc28be6e9890df4c84b included this change:

```diff
@@ -627,7 +628,7 @@ Handshake::read_port() {
   m_readBuffer.read_8();
 
   if (length == 2)
-    manager->dht_manager()->add_node(m_address.c_sockaddr(), m_readBuffer.peek_16());
+    manager->dht_manager()->add_node(m_address.get(), m_readBuffer.read_16());
 
   m_readBuffer.consume(length);
   return true;
```

This is probably the cause of #256. Debug logging shows the value of `ProtocolBuffer::remaining()` going to `65534`, i.e., `-2`, since `m_position` advances an additional 2 beyond `m_end`.

### DIFF
--- a/src/protocol/handshake.cc
+++ b/src/protocol/handshake.cc
@@ -630,7 +630,7 @@ Handshake::read_port() {
   m_readBuffer.read_8();
 
   if (length == 2)
-    manager->dht_manager()->add_node(m_address.get(), m_readBuffer.read_16());
+    manager->dht_manager()->add_node(m_address.get(), m_readBuffer.peek_16());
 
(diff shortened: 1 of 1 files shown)


rakshasa:
Since you can trigger the issue, add a log entry to the if block where dht add_node is done and two temporary log entries sandwiching the consume call.

Once you verify that it happens due to the dht add_node call remove the two temporary logging events.

Issue #253, opened 2024-10-03
Replace legacy sync* builtins with modern atomic* builtins: fixes build on 32-bit platforms


barracuda156:
@rakshasa Could you please take a look?

The issue is that old `sync*` builtins require 8-byte atomics support in hardware. That leaves 32-bit platforms (arm, ppc, mips) broken. What will work for both 64- and 32-bit platforms alike is usage of modern `atomic*` builtins. 32-bit archs will just need an explicit linking to `libatomic`.

This a patch which is confirmed to fix compilation on `ppc`, but maybe you can suggest a better solution or verify/improve upon this one?

https://github.com/macports/macports-ports/blob/4ac3291874cad2051b9cc1ba08cadd62643f92f5/net/libtorrent-devel/files/0001-Fix-atomics-for-modern-gcc.patch

stickz:
@barracuda156 Could you submit a pull request for this patch? I'd like to take responsibility for testing this change and getting it incorporated into the project. I'll request a review from @rakshasa when the time is right and get it merged for you.

rakshasa:
I'd rather replace them with `std::atomic`, as that will avoid any compiler dependency.

barracuda156:
@stickz Sorry, I am away from the testing hardware at the moment, so can’t do much until I am back.

@rakshasa If you can handle this, it will be much better, since I am not too sure that my fix is the right one.

(FWIW, we will still depend on `libatomic` on some 32-bit archs, AFAIU, as long as 8-byte atomics is needed in some form.)

rakshasa:
All `__sync_` calls removed, so `libatomic` should no longer be required.

barracuda156:
@rakshasa Thank you very much, I will try it out.

barracuda156:
@rakshasa Everything compiles fine now for me, thanks again.

PR #265, opened 2024-12-01
Convert __sync* built-ins to std::atomic

kannibalox:
I suspect these changes mean that `lt_cacheline_aligned` is no longer required for some of the objects, but left them alone since I wasn't 100% on that.

Addresses https://github.com/rakshasa/libtorrent/issues/253

Notable changes:
- The thread_base::*_main_polling() functions didn't appear to be used outside of a test, so I removed them entirely
- thread_base's pthread_mutex_t was changed to std::mutex
- instrumentation.cc was changed to log lines indicating when libtorrent wasn't compiled with it enabled
- If `thread_base::m_global` was simply changed to use std::atomic/std::mutex, it couldn't be initialized as a protected static member outside of the class in C++14 due to not being able to guarantee copy elision for std::atomic's construction. C++17 fixes this, but in the meantime I broke it out into two separate members since we shouldn't strictly need to `lt_cacheline_aligned` them together anymore.


src/torrent/utils/thread_interrupt.h
```diff
@@ -37,6 +37,8 @@
 #ifndef LIBTORRENT_UTILS_THREAD_INTERRUPT_H
 #define LIBTORRENT_UTILS_THREAD_INTERRUPT_H
 
+#include "config.h"
```
rakshasa, line 40: This should never be in a header file.

rakshasa:
Check fix/atomic-types before making any changes.

src/torrent/utils/thread_interrupt.h
kannibalox: I have no quarrel with that, but it throws off clang-tidy since it can't infer what `lt_cacheline_aligned` is.

kannibalox:
I have no problem with this getting closed in favor of your work, otherwise I'll wait until it's merged into master and compare/contrast at that point.

src/torrent/utils/thread_base.h
```diff
@@ -144,46 +136,29 @@ thread_base::send_event_signal(unsigned int index, bool do_interrupt) {
 
 inline void
 thread_base::acquire_global_lock() {
-  __sync_add_and_fetch(&thread_base::m_global.waiting, 1);
-  pthread_mutex_lock(&thread_base::m_global.lock);
-  __sync_sub_and_fetch(&thread_base::m_global.waiting, 1);
+  ++m_waiting;
+  m_globalLock.lock();
+  --m_waiting;
```
rakshasa, line 141: This should be using std::counting_semaphore.

rakshasa:
I'd rather do the conversion myself as they need careful consideration and there are some improvements I'd like to do.

PR #268, opened 2024-12-06
Fix thread unit tests

kannibalox:
Convert m_poking back to a bool due to test_and_set having different return values than compare_exchange. Without this test_thread_base fails:
```
test_thread_base.cc:61:Assertion
Test name: test_thread_base::test_lifecycle
assertion failed
- Expression: wait_for_true(std::bind(&test_thread::is_test_state, thread, test
_thread::TEST_PRE_STOP))

test_thread_base.cc:130:Assertion
Test name: test_thread_base::test_interrupt
assertion failed
- Expression: wait_for_true(std::bind(&test_thread::is_not_test_flags, thread, 
test_thread::test_flag_do_work))

Failures !!!
Run: 27   Failure total: 2   Failures: 2   Errors: 0
FAIL: LibTorrent_Test_Torrent_Utils
```
The first failure is fixed by the test_thread.h change, the second by the m_poking change.

### DIFF
--- a/src/torrent/utils/thread_interrupt.cc
+++ b/src/torrent/utils/thread_interrupt.cc
@@ -27,7 +27,8 @@ thread_interrupt::~thread_interrupt() {
 
 bool
 thread_interrupt::poke() {
-  if (m_poking.test_and_set())
+  bool expected = false;
+  if (!m_other->m_poking.compare_exchange_strong(expected, true))
     return true;
 
   int result = ::send(m_fileDesc, "a", 1, 0);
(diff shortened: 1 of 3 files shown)


src/torrent/utils/thread_interrupt.cc
```diff
@@ -73,6 +70,8 @@ thread_interrupt::event_read() {
     throw internal_error("Invalid result reading from thread_interrupt socket.");
 
   instrumentation_update(INSTRUMENTATION_POLLING_INTERRUPT_READ_EVENT, 1);
+
+  m_poking = false;
```
kannibalox, line 74: I didn't dig into exactly why, but moving this back down is indeed required for the tests to succeed.

PR #274, opened 2024-12-13
Clean up rak::functional's last methods

kannibalox:
This is the final cleanup for functional.h. `slot_list_call` doesn't have a clear equivalent, so I just simplified it and left it in the same file.

There is a logic change in HashQueue::remove(), so that it does the removal in a single pass rather than erasing then re-finding individual iterators: https://github.com/rakshasa/libtorrent/compare/master...kannibalox:libtorrent:feature/rakfn-to-lambda#diff-0445aee53785eeeccbb508f105cb7fc2665fa18637ee474c9317dc150c2a9546L118-L151

### DIFF
--- a/rak/functional.h
+++ b/rak/functional.h
@@ -5,11 +5,35 @@
 
 namespace rak {
 
-template <typename Container, typename... Args>
+template <typename Container>
 inline void
-slot_list_call(const Container& slot_list, Args&&... args) {
-  for (const auto& slot : slot_list) {
-    slot(std::forward<Args>(args)...);
+slot_list_call(const Container& slot_list) {
+  if (slot_list.empty())
+    return;
+
+  typename Container::const_iterator first = slot_list.begin();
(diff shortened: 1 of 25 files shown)


rak/functional.h
```diff
@@ -1,183 +1,18 @@
...
-  typename Container::const_iterator first = slot_list.begin();
-  typename Container::const_iterator next = slot_list.begin();
-
-  while (++next != slot_list.end()) {
-    (*first)(arg1);
-    first = next;
+slot_list_call(const Container& slot_list, Args&&... args) {
+  for (const auto& slot : slot_list) {
+    slot(std::forward<Args>(args)...);
   }
```
rakshasa, line 13: When calling slot the list entry can remove itself.

This is why the container must be a std::list (or equivalently container type) and we need to have the next iterator entry before calling slot.

kannibalox: Gotcha, reverted to original code.

PR #277, opened 2024-12-18
remove most ptr_fun

neheb:
Removed in C++17

### DIFF
--- a/src/protocol/handshake_manager.cc
+++ b/src/protocol/handshake_manager.cc
@@ -31,14 +31,6 @@ namespace torrent {
 
 ProtocolExtension HandshakeManager::DefaultExtensions = ProtocolExtension::make_default();
 
-inline void
-handshake_manager_delete_handshake(Handshake* h) {
-  h->deactivate_connection();
-  h->destroy_connection();
-
-  delete h;
-}
-
 HandshakeManager::size_type
 HandshakeManager::size_info(DownloadMain* info) const {
   return std::count_if(base_type::begin(), base_type::end(), [info](Handshake* h) { return info == h->download(); });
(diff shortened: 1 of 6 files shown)


src/protocol/handshake_manager.cc
```diff
@@ -60,15 +60,10 @@ HandshakeManager::erase(Handshake* handshake) {
...
-  bool operator () (const rak::socket_address* sa1, const Handshake* p2) const {
-    return p2->peer_info() != NULL && *sa1 == *rak::socket_address::cast_from(p2->peer_info()->socket_address());
-  }
-};
-
 bool
 HandshakeManager::find(const rak::socket_address& sa) {
-  return std::find_if(base_type::begin(), base_type::end(), std::bind1st(handshake_manager_equal(), &sa)) != base_type::end();
+  auto f = [&sa](const Handshake* p2) { return p2->peer_info() && sa == *rak::socket_address::cast_from(p2->peer_info()->socket_address()); };
+  return std::any_of(base_type::begin(), base_type::end(), f);
```
rakshasa, line 66: Wrong style, should not use an intermediate function variable.

src/protocol/handshake_manager.cc
```diff
@@ -77,7 +72,7 @@ HandshakeManager::erase_download(DownloadMain* info) {
     return info != h->download();
   });
 
-  std::for_each(split, base_type::end(), std::ptr_fun(&handshake_manager_delete_handshake));
+  std::for_each(split, base_type::end(), &handshake_manager_delete_handshake);
```
rakshasa, line 75: Use lambda.

src/protocol/handshake_manager.cc
neheb: It’s pretty long though

rakshasa: It should be multiple lines, and is what I find to be the best style for these cases.

src/protocol/handshake_manager.cc
neheb: Done.

src/protocol/handshake_manager.cc
neheb: Done. clang-format seems happy.

PR #278, opened 2024-12-18
unique_ptr

neheb:


rakshasa:
Every PR I have to manually check out locally and run CI against, please make any cleanup or c++ migration PR's more than just a couple of lines.

neheb:
noted. I'll make a big one soon.

Are you a fan of migrating to std::mem_fn in place of lambdas?

PR #281, opened 2024-12-19
convert structs with single operator() to funcs

neheb:


src/download/download_constructor.cc
```diff
@@ -57,28 +57,19 @@
...
 
-struct download_constructor_is_single_path {
-  bool operator () (Object::map_type::const_reference v) const {
-    return
-      std::strncmp(v.first.c_str(), "name.", sizeof("name.") - 1) == 0 &&
-      v.second.is_string();
-  }
+bool
+download_constructor_is_single_path(Object::map_type::const_reference v) {
+  return v.first.find("name.") == 0 && v.second.is_list();
```
rakshasa, line 62: This should be `is_string`.

src/torrent/tracker_list.cc
```diff
@@ -84,14 +84,14 @@ TrackerList::has_active_not_scrape_in_group(uint32_t group) const {
...
-  bool operator () (const TrackerList::value_type& value) const { return value->is_usable(); }
+bool
+tracker_usable_t(const TrackerList::value_type& value) {
+  return value->is_usable();
 };
 
 bool
 TrackerList::has_usable() const {
-  return std::find_if(begin(), end(), tracker_usable_t()) != end();
+  return std::any_of(begin(), end(), tracker_usable_t);
```
rakshasa, line 94: Simple functions like this should be using lambda functions.

src/download/download_constructor.cc
neheb: Huh. I wonder how that slipped in.

src/torrent/tracker_list.cc
neheb: Will do. I avoided doing so as to not change it too much.

src/download/download_main.cc
```diff
@@ -424,11 +421,10 @@ DownloadMain::do_peer_exchange() {
     // Create the new m_ut_pex_list by removing any 'removed'
     // addresses from the original list and then adding the new
     // addresses.
-    m_ut_pex_list.erase(std::set_difference(m_ut_pex_list.begin(), m_ut_pex_list.end(), removed.begin(), removed.end(),
-                                            m_ut_pex_list.begin(), SocketAddressCompact_less()), m_ut_pex_list.end());
+    m_ut_pex_list.erase(std::set_difference(m_ut_pex_list.begin(), m_ut_pex_list.end(), removed.begin(), removed.end(), m_ut_pex_list.begin(), SocketAddressCompact_less), m_ut_pex_list.end());
```
rakshasa, line 424: The `set_difference` calls are so long they should be multiline, with each begin/end and ftor on separate lines.

neheb: This is the best I could get clang-format to do.

neheb:
Switched to std::mem_fn instead of lambdas where shorter.

Issue #284, opened 2024-12-26
FreeBSD: 0.15.0 fails to build


nunotexbsd:
FreeBSD 15
Clang 19.1.5
libtorrent [port](https://www.freshports.org/net-p2p/libtorrent/)

Any clues?
Full log bellow from poudriere packager.

Thanks

```
--- throttle.lo ---
In file included from throttle.cc:43:
In file included from ./../net/throttle_internal.h:41:
../../rak/priority_queue_default.h:56:7: warning: '~priority_item' has a non-throwing exception specification but can still throw [-Wexceptions]
   56 |       throw torrent::internal_error("priority_item::~priority_item() called on a queued item.");
      |       ^
../../rak/priority_queue_default.h:54:3: note: destructor has a implicit non-throwing exception specification
   54 |   ~priority_item() {
      |   ^
--- rate.lo ---
In file included from rate.cc:39:
In file included from ./../globals.h:5:
../../rak/priority_queue_default.h:56:7: warning: '~priority_item' has a non-throwing exception specification but can still throw [-Wexceptions]
   56 |       throw torrent::internal_error("priority_item::~priority_item() called on a queued item.");
      |       ^
../../rak/priority_queue_default.h:54:3: note: destructor has a implicit non-throwing exception specification
   54 |   ~priority_item() {
      |   ^
--- poll_kqueue.lo ---
poll_kqueue.cc:284:41: error: no member named 'equal' in namespace 'rak'
  284 |                                    rak::equal(event, rak::mem_ref(&kevent::udata))) - m_changes;
      |                                    ~~~~~^
poll_kqueue.cc:284:59: error: no member named 'mem_ref' in namespace 'rak'
  284 |                                    rak::equal(event, rak::mem_ref(&kevent::udata))) - m_changes;
      |                                                      ~~~~~^
--- poll_select.lo ---
poll_select.cc:127:5: warning: '~PollSelect' has a non-throwing exception specification but can still throw [-Wexceptions]
  127 |     throw internal_error("PollSelect::~PollSelect() called but the sets are not empty");
      |     ^
poll_select.cc:120:13: note: destructor has a implicit non-throwing exception specification
--- poll_kqueue.lo ---
poll_kqueue.cc:310:41: error: no member named 'equal' in namespace 'rak'
  310 |                                    rak::equal(event, rak::mem_ref(&kevent::udata))) - m_changes;
      |                                    ~~~~~^
[...]

PR #286, opened 2024-12-27
Refactor deleted rak::mem_ref reference

kannibalox:
CC @nunotexbsd

Fixes https://github.com/rakshasa/libtorrent/issues/284

### DIFF
--- a/src/torrent/poll_kqueue.cc
+++ b/src/torrent/poll_kqueue.cc
@@ -279,7 +279,7 @@ PollKQueue::close(Event* event) {
       itr->udata = NULL;
 
   m_changedEvents = std::remove_if(m_changes, m_changes + m_changedEvents,
-                                   rak::equal(event, rak::mem_ref(&kevent::udata))) - m_changes;
+                                   [event](struct kevent& e) { return event == e.udata; }) - m_changes;
 }
 
 void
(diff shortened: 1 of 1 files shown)


nunotexbsd:
Build fixed.
Thanks for quick support.

chenrui333:
same pr as #285

src/torrent/poll_kqueue.cc
```diff
@@ -279,7 +279,7 @@ PollKQueue::close(Event* event) {
       itr->udata = NULL;
 
   m_changedEvents = std::remove_if(m_changes, m_changes + m_changedEvents,
-                                   rak::equal(event, rak::mem_ref(&kevent::udata))) - m_changes;
+                                   [event](struct kevent& e) { return event == e.udata; }) - m_changes;
```
neheb, line 282: should be using a const reference. struct is also optional in C++.

I wonder if adding a bool operatror== would allow changing remove_if to remove.

neheb: or std::men_fn. Whichever is shorter.

kannibalox:
Closing in favor of https://github.com/rakshasa/libtorrent/pull/285, essentially the same PR but opened before mine.

src/torrent/poll_kqueue.cc
kannibalox: The new PR uses const, however it'd probably be better to avoid even simple logic changes in these hotfix-ish PRs.

PR #289, opened 2025-01-12
Bump c++ standard to 17

stickz:
Compiler warnings are resolved: `warning: #import is a deprecated GCC extension [-Wdeprecated]`

Resolves crash when compiling libtorrent with `-march=native`. https://github.com/rakshasa/rtorrent/issues/1319#issuecomment-2580534716


neheb (approved):

neheb:
Would love to replace this pthread stuff with https://en.cppreference.com/w/cpp/thread/scoped_lock

PR #299, opened 2025-01-20
more clang-tidy fixes

neheb:


neheb:
That m_peerChunks unique_ptr looks like it should be a vector IMO.

rakshasa:
This is too much to review for no real immediate benefit.

While I want code converted as new PR's touch upon them, going through this much just for syntactical cleanup is not worth it.

neheb:
I could reduce the commit count if you want.

rakshasa:
Won't change the fact that most of those changes are not worth the time reviewing.

rakshasa:
Please don't spend more time on a PR that won't be merged. These changes do not matter enough for me to review over >100 files.

neheb:
Removed 2 commits, one of which was the huge default one.

src/net/protocol_buffer.h
```diff
@@ -141,10 +141,7 @@ inline bool
 ProtocolBuffer<tmpl_size>::consume(difference_type v) {
   m_position += v;
 
-  if (remaining())
-    return false;
-
-  return true; 
+  return !static_cast<bool>(remaining());
```
rakshasa, line 144: This makes it more difficult to read.

src/torrent/data/block.h
```diff
@@ -163,22 +163,16 @@ Block::Block() :
 
 inline BlockTransfer*
 Block::find(const PeerInfo* p) {
-  BlockTransfer* transfer;
-
-  if ((transfer = find_queued(p)) != NULL)
+  if (auto transfer = find_queued(p))
```
rakshasa, line 166: The style we use is to always check against nullptr.

src/torrent/tracker.h
```diff
@@ -159,7 +159,7 @@ class LIBTORRENT_EXPORT Tracker {
 
 inline bool
 Tracker::can_request_state() const {
-  return !(is_busy() && latest_event() != EVENT_SCRAPE) && is_usable();
+  return (!is_busy() || latest_event() == EVENT_SCRAPE) && is_usable();
```
rakshasa, line 162: The way these are written is intentional, it tells the reader what the meaning of the check is.

These should not be reduced.

src/torrent/data/block.h
neheb: right. this is a C++17 feature.

Would you prefer

```
auto transfer = find_queue(p);
if (transfer != nullptr)
```

src/net/protocol_buffer.h
neheb: How about removing the cast and just returning

```
return !remaining();
```

rakshasa: No, the style I used originally makes it very clear what is happening at a glance, while you have to think more with the compact style you are suggesting.

rakshasa:
Close this issue and open a new PR so the checks can be run again.

PR #307, opened 2025-02-04
add download sequentially

Auska:


rakshasa:
This will make a lot of people very angry, so never merging a straight-up sequential sequencer.

Either make it an option to partially download a set amount of the start of files. (with hard-capped limit)

Or to improve performance (if that's what you want), improve the logic that decides when to trigger randomize.

PR #309, opened 2025-02-06
some clang-tidy

neheb:


rakshasa:
Again these are not changes that materially improve the code, and in some places actually makes it more obtuse.

E.g. expanding compact if/else statements to multiple lines and adding operators to Chunks when the explicit member function call is more descriptive.

src/torrent/bitfield.h
```diff
@@ -50,8 +50,10 @@ class LIBTORRENT_EXPORT Bitfield {
   typedef value_type*           iterator;
   typedef const value_type*     const_iterator;
 
-  Bitfield() : m_size(0), m_set(0), m_data(NULL)    {}
+  Bitfield() = default;
   ~Bitfield()                                       { clear(); }
+  Bitfield(const Bitfield&) = delete;
+  Bitfield& operator=(const Bitfield&) = delete;
```
rakshasa, line 56: Make a separate PR with just these changes to ctor, deleted operators and default member variable values.

Also respect empty newlines, don't pack different categories of member functions together like that.

PR #318, opened 2025-02-11
replace usleep with condition_variable

neheb:
It's more efficient.

### DIFF
--- a/src/torrent/utils/thread_base.cc
+++ b/src/torrent/utils/thread_base.cc
@@ -59,8 +59,9 @@ thread_base::stop_thread_wait() {
 
   release_global_lock();
 
-  while (!is_inactive()) {
-    usleep(1000);
+  {
+    auto lock = std::unique_lock(m_global.mutex);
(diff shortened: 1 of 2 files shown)


src/torrent/utils/thread_base.cc
```diff
@@ -59,8 +59,9 @@ thread_base::stop_thread_wait() {
 
   release_global_lock();
 
-  while (!is_inactive()) {
-    usleep(1000);
+  {
+    auto lock = std::unique_lock(m_global.mutex);
+    m_global.cv.wait(lock, [this] { return !is_inactive(); });
```
rakshasa, line 64: This is wrong, you only ever acquire the global lock through acquire_global_lock() as the waiting value must be incremented.

Which means this cv implementation is incorrect.

rakshasa, line 64: This is wrong, you only ever acquire the global lock through acquire_global_lock() as the waiting value must be incremented.

Which means this cv implementation is incorrect.

PR #327, opened 2025-03-04
replace usleep with pthread_join

uis246:


uis246:
I didn't do test-compile because I did not want to fight with autoconf.
Alternative to #318.

uis246:
`is_active()` may be excessive here, because pthread_join manual says joining dead pthread will cause instant return.

rakshasa:
Yeah, assert would be correct.

uis246:
Done

uis246:
`is_inactive()` seems to be never outside of thread_base and tests.

rakshasa:
```
utils/thread_base.cc:64:3: error: ‘assert’ was not declared in this scope
   64 |   assert(!is_inactive());
      |   ^~~~~~
utils/thread_base.cc:12:1: note: ‘assert’ is defined in header ‘<cassert>’; did you forget to ‘#include <cassert>’?
   11 | #include "utils/instrumentation.h"
  +++ |+#include <cassert>
```

rakshasa:
Merged updated version.

uis246:
Right, I even forgot about it. Sorry. Only got time in last few days.

PR #331, opened 2025-03-12
Fix invalid memory accesses on exit

jmakovicka:
rtorrent now segfaults in `Manager::remove_controller` after adding a new magnet link when nothing is being downloaded yet.


neheb:
would be great to transition to C++20 to get std::erase_if eventually.

PR #344, opened 2025-03-31
fix compilation with g++8

neheb:
Some platforms still use this old version.

### DIFF
--- a/src/net/udns/udns_resolver.c
+++ b/src/net/udns/udns_resolver.c
@@ -1241,7 +1241,7 @@ struct dns_resolve_data {
   void *dnsrd_result;
 };
 
-static void dns_resolve_cb([[maybe_unused]] struct dns_ctx *ctx, void *result, void *data) {
+static void dns_resolve_cb(__attribute__((unused)) struct dns_ctx *ctx, void *result, void *data) {
   struct dns_resolve_data *d = (struct dns_resolve_data *)data;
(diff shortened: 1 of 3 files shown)


src/torrent/utils/thread.h
```diff
@@ -9,6 +9,8 @@
 #include <sys/types.h>
 
 #include <torrent/common.h>
+#include <torrent/net/resolver.h>
+#include <torrent/poll.h>
```
rakshasa, line 13: Try removing the `nullptr` from `m_poll{nullptr};` and the other, those headers should not be included here.

neheb: That indeed fixed it.

src/net/udns/udns_resolver.c
```diff
@@ -1241,7 +1241,7 @@ struct dns_resolve_data {
   void *dnsrd_result;
 };
 
-static void dns_resolve_cb([[maybe_unused]] struct dns_ctx *ctx, void *result, void *data) {
+static void dns_resolve_cb(__attribute__((unused)) struct dns_ctx *ctx, void *result, void *data) {
```
rakshasa, line 1244: Not going back to using `__attribute__`.

neheb: This is actually an issue that C compilers don’t support attributes until C23.

PR #348, opened 2025-04-02
Replace usleep with pthread_join.

rakshasa:


uis246:
Is there a reason to not remove unistd.h include?

PR #357, opened 2025-04-07
replace some binds with lambdas

neheb:


src/torrent/peer/client_list.cc
```diff
@@ -302,8 +302,10 @@ ClientList::retrieve_id(ClientInfo* dest, const HashString& id) const {
     return false;
   }
 
-  const_iterator itr = std::find_if(begin() + 1, end(), std::bind(&ClientInfo::intersects, *dest, std::placeholders::_1));
-
+  auto itr = std::find_if(
+    std::next(begin()),
+    end(),
+    [dest](const auto& client) { return ClientInfo::intersects(*dest, client); });
```
rakshasa, line 308: Make these two:

```
  auto itr = std::find_if(std::next(begin()), end(), [dest](const auto& client) {
      return ClientInfo::intersects(*dest, client);
    });
```

(indent may not be correct)

neheb: Done

PR #375, opened 2025-04-15
some compiler fixes

neheb:


src/net/udns/udns_resolver.c
```diff
@@ -887,7 +887,7 @@ dns_submit_dn(struct dns_ctx *ctx,
   SETCTXOPEN(ctx);
   dns_assert_ctx(ctx);
 
-  q = (struct dns_query *)calloc(sizeof(*q), 1);
+  q = (struct dns_query *)calloc(1, sizeof(*q));
```
rakshasa, line 890: This should probably also be fixed in the upstream udns code.

neheb: oh weird that this isn't available as an external dependency.

PR #383, opened 2025-04-17
add extra CI

neheb:
Add extra platforms to make sure they don't regress.

Commented out OpenBSD as it doesn't work currently. Also comment out tests where they currently fail. OmniOS has no cppunit package. BSD tests fail on thread tests.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -3,36 +3,117 @@ name: Unit tests
 on: pull_request
 
 jobs:
-
   unit-tests:
     runs-on: ubuntu-22.04
-
     steps:
-    - name: Update Packages
+    - name: Prepare
       run: |
         sudo apt-get update
-    - name: Install Dependencies
-      run: |
         sudo apt-get install -y libcppunit-dev
     - uses: actions/checkout@v4
-      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -36,3 +36,106 @@ jobs:
...
+      - name: Install packages
+        run: |
+          brew install automake cppunit libtool
+
+      - name: Compile and Test
+        run: |
+          autoreconf -fiv
+          ./configure
+          make
+          make check
```
rakshasa, line 53: We should not have different steps from the original.

neheb: Done

.github/workflows/unit-tests.yml
```diff
@@ -3,36 +3,124 @@ name: Unit tests
...
     - name: Run Unit Tests
       run: |
         make check
+  MacOS:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install packages
+        run: |
+          brew install automake cppunit libtool
```
rakshasa, line 32: Install dependencies before checkout so that cached images can be used.

PR #384, opened 2025-04-17
unique_ptr conversions

neheb:


neheb:
oh it turns out this fill is not even needed.

when unique_ptr 's type is T[], it calls memset.

PR #385, opened 2025-04-17
clang-tidy fixes

neheb:


src/torrent/common.h
```diff
@@ -81,7 +82,6 @@ class Resolver;
 
 namespace tracker {
 class DhtController;
-class Manager;
```
rakshasa, line 84: This is wrong, tracker::Manager is supposed to be there and torrent::Manager is internal to libtorrent.

neheb: removed.

src/torrent/tracker_list.cc
```diff
@@ -357,16 +357,16 @@ TrackerList::size_group() const {
...
-  if (itr == end() || (*itr).group() != group)
+  if (first == end() || first->group() != group)
     return;
 
-  while (++itr != end() && (*itr).group() == group) {
-    std::iter_swap(itr, prev);
-    prev = itr;
-  }
+  auto last = first;
+  while (last != end() && last->group() != group)
```
rakshasa, line 366: Wrong comparison.

neheb: fixed.

PR #387, opened 2025-04-18
clang-tidy: default initialize members

neheb:
Found with modernize-use-default-member-init


src/dht/dht_router.cc
```diff
@@ -28,9 +28,7 @@ HashString DhtRouter::zero_id;
 
 DhtRouter::DhtRouter(const Object& cache, const rak::socket_address* sa) :
   DhtNode(zero_id, sa),  // actual ID is set later
-  m_server(this),
```
rakshasa, line 31: Requires `this`/

neheb: interesting... Wonder why this was removed.

PR #390, opened 2025-04-20
CI: test big endian platform as well

neheb:
Slower than the others. Probably should be done as rtorrent is used on big endian platforms as well (mostly routers).


rakshasa:
72 minutes?

Sorry, but that will need to be run in some other way, not blocking PRs.

Perhaps run it after a merge to master so it generates an issue or notification.

PR #396, opened 2025-04-21
various clang-tidy fixes

neheb:


rakshasa:
```
this_thread::scheduler()->wait_for_ceil_seconds(&m_delay_scrape, 10s);
this_thread::poll()->remove_read(this);
```

The above are new ways of interacting with poll and a replacement for priority_queue.

PRs (of moderate size for ease of review) that migrates to these would be more useful than language modernization / cleanup.

PR #407, opened 2025-05-02
Refactor/remove priority queue

rakshasa:


src/thread_main.cc
```diff
@@ -92,22 +92,12 @@ std::chrono::microseconds
...
-
-  if (m_slot_next_timeout) {
-    // TODO: Change slot to chrono.
-    auto slot_timeout = std::max<uint64_t>(m_slot_next_timeout(), 0);
-
-    timeout = std::min(timeout, std::chrono::microseconds(slot_timeout));
-  }
-
-  return timeout;
+  return std::min(std::chrono::microseconds(task_timeout), std::chrono::microseconds(10s));
```
neheb, line 100: I think you can do std::min<std::chrono::microseconds>. Slightly shorter.

rakshasa: Doesn't work since there's no implicit conversion of `task_timeout`.

PR #416, opened 2025-05-05
replace aligned_alloc with posix_memalign

neheb:
Seemed like a good idea, but it's throwing AddressSantizer errors when size is not a multiple of the cacheline size. It's also unsupported by Android API < 28.

ping @biswa96


neheb:
interesting. this fixes some tests with my meson build but unaffected here.

PR #419, opened 2025-05-07
Removed lt_cacheline_aligned.

rakshasa:


neheb (approved):

neheb:
for the remaining macro in meson, I use

```
if cc.has_header_symbol('new', 'std::hardware_constructive_interference_size')
  cdata.set('LT_SMP_CACHE_BYTES', 'std::hardware_constructive_interference_size')
else
  cdata.set('LT_SMP_CACHE_BYTES', '128')
endif
```

not sure if that's the intention behind it. Required C++20 for the true statement.

rakshasa:
https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size

Since c++17, so we can use it. Though it might not be available with some older compilers.

neheb:
right. I don't think GCC8 has it for example.

PR #420, opened 2025-05-07
simplify cacheline_allocator with C++16

neheb:
new and delete support aligned allocations now. No need for posix_memalign. AddressSanitizer also doesn't complain.

Kept the class to just allocate() and deallocate(). Nothing else is needed.


neheb:
tested all the way back to GCC7. Compiles.

rakshasa:
Should be also be moved to `torrent/utils`.

PR #421, opened 2025-05-09
sha1: remove deprecated API usage

neheb:


rakshasa:
```
CI results:

failed:    ci-default
failed:    ci-low-diskspace
failed:    ci-nat
failed:    ci-throttle
failed:    ci-torrents
```

rakshasa:
Seems most likely it's happening during handshake.

neheb:
Now I wonder if it’s the sha1 or rc4 patch causing the issue

rakshasa:
Cherry-picked, and they both fail individually.

With just master it still works.

neheb:
switched to the _ex variants. They don't call _reset it seems.

neheb:
removed rc4 change. I have an idea for it. Will split it up into its own thing.

rakshasa:
We hash large blocks not <128 bytes, so that's not a performance issue.

PR #424, opened 2025-05-10
rc4: remove deprecated API usage 

neheb:


src/protocol/encryption_info.h
```diff
@@ -37,32 +37,34 @@
...
 
 private:
   bool                m_encrypted{false};
   bool                m_obfuscated{false};
   bool                m_decryptValid{false};
 
-  RC4                 m_encrypt;
-  RC4                 m_decrypt;
+  std::unique_ptr<RC4> m_encrypt;
+  std::unique_ptr<RC4> m_decrypt;
```
rakshasa, line 67: Using pointers has performance costs, so unless there's a reason for it it is better to embed the value itself.

It's either to reduce header includes, or when it affects the code itself simpler, but here and below you are not gaining either.

rakshasa, line 67: Using pointers has performance costs, so unless there's a reason for it it is better to embed the value itself.

It's either to reduce header includes, or when it affects the code itself simpler, but here and below you are not gaining either.

If you can't copy RC4, then make that explicit by deleting copy operators and changing `set_*`.

neheb: Hrm OK

neheb: No good way to do so:

```
error: use of deleted function 'torrent::EncryptionInfo& torrent::EncryptionInfo::operator=(const torrent::EncryptionInfo&)'
  129 |   m_encryption = *encryptionInfo;
```

better to keep as is.

rakshasa:
```
failed:    ci-default
failed:    ci-low-diskspace
failed:    ci-nat
failed:    ci-throttle
failed:    ci-torrents
```

Still failing.

src/protocol/encryption_info.h
```diff
@@ -53,8 +55,8 @@ class EncryptionInfo {
   bool                decrypt_valid() const             { return m_decryptValid; }
 
   void                set_obfuscated()                  { m_obfuscated = true; m_encrypted = m_decryptValid = false; }
-  void                set_encrypt(const RC4& encrypt)   { m_encrypt = encrypt; m_encrypted = m_obfuscated = true; }
-  void                set_decrypt(const RC4& decrypt)   { m_decrypt = decrypt; m_decryptValid = true; }
+  void                set_encrypt(RC4 encrypt)          { m_encrypt = std::move(encrypt); m_encrypted = m_obfuscated = true; }
+  void                set_decrypt(RC4 decrypt)          { m_decrypt = std::move(decrypt); m_decryptValid = true; }
```
rakshasa, line 59: Seems these should be `set_*(RC4&& *crypt)`.

neheb: hmm? they're still moved.

rakshasa: By using rvalue reference you enforce the use of `std::move`.

neheb: sure. although there's no advantage here. As I mentioned earlier, it's hard to remove copies without pointers.

reading https://clang.llvm.org/extra/clang-tidy/checks/modernize/pass-by-value.html , before this PR, the RC4 call uses the copy constructor because of const ref. When it takes a value parameter, the compiler is free to choose copy or move. Although the way I modified it, it's move.

Although reading OpenSSL code is convincing me custom copy or move constructors need to be written...

rakshasa: My point was that using `RC&&` makes it so there's less chance that that any future changes to the code misses the need requirement that RC4 is never copied.

It's about using how you write the code to express what your intentions are, and here we want to say RC4 is never copied.

That's why imo it should be using rvalue reference, not value, even if the latter results in the correct output.

neheb: Done.

neheb:
I was able to get rid of copies with

```
diff --git a/src/protocol/peer_connection_base.cc b/src/protocol/peer_connection_base.cc
index 28d560d54e7b..e5a5e7ca87df 100644
--- a/src/protocol/peer_connection_base.cc
+++ b/src/protocol/peer_connection_base.cc
@@ -126,7 +126,7 @@ PeerConnectionBase::initialize(DownloadMain* download, PeerInfo* peerInfo, Socke
   m_peerInfo = peerInfo;
   m_download = download;
 
-  m_encryption = *encryptionInfo;
+  m_encryption = encryptionInfo;
   m_extensions = extensions;
 
   m_extensions->set_connection(this);
@@ -535,7 +535,7 @@ PeerConnectionBase::down_chunk() {
     data.second = read_stream_throws(data.first, data.second);
 
     if (is_encrypted())
-      m_encryption.decrypt(data.first, data.second);
+      m_encryption->decrypt(data.first, data.second);
 
     bytesTransfered += data.second;
 
@@ -581,7 +581,7 @@ PeerConnectionBase::down_chunk_skip() {
   throttle->node_used(m_peerChunks.download_throttle(), length);
 
   if (is_encrypted())
-    m_encryption.decrypt(m_nullBuffer, length);
+    m_encryption->decrypt(m_nullBuffer, length);
 
   if (down_chunk_skip_process(m_nullBuffer, length) != length)
     throw internal_error("PeerConnectionBase::down_chunk_skip() down_chunk_skip_process(m_nullBuffer, length) != length.");
@@ -695,7 +695,7 @@ PeerConnectionBase::down_extension() {
     m_down->throttle()->node_used_unthrottled(bytes);
     
     if (is_encrypted())
-      m_encryption.decrypt(m_extensions->read_position(), bytes);
+      m_encryption->decrypt(m_extensions->read_position(), bytes);
 
     m_extensions->read_move(bytes);
   }
@@ -733,7 +733,7 @@ PeerConnectionBase::up_chunk_encrypt(uint32_t quota) {
   }
 
   m_upChunk.chunk()->to_buffer(m_encryptBuffer->end(), m_upPiece.offset() + m_encryptBuffer->remaining(), quota);
-  m_encryption.encrypt(m_encryptBuffer->end(), quota);
+  m_encryption->encrypt(m_encryptBuffer->end(), quota);
   m_encryptBuffer->move_end(quota);
 
   return m_encryptBuffer->remaining();
@@ -794,12 +794,12 @@ bool
 PeerConnectionBase::up_extension() {
   if (m_extensionOffset == extension_must_encrypt) {
     if (m_extensionMessage.owned()) {
-      m_encryption.encrypt(m_extensionMessage.data(), m_extensionMessage.length());
+      m_encryption->encrypt(m_extensionMessage.data(), m_extensionMessage.length());
 
     } else {
       auto buffer = new char[m_extensionMessage.length()];
 
-      m_encryption.encrypt(m_extensionMessage.data(), buffer, m_extensionMessage.length());
+      m_encryption->encrypt(m_extensionMessage.data(), buffer, m_extensionMessage.length());
       m_extensionMessage.set(buffer, buffer + m_extensionMessage.length(), true);
     }
 
diff --git a/src/protocol/peer_connection_base.h b/src/protocol/peer_connection_base.h
index 7f9de5f2a06f..4e37a2f2b5b8 100644
--- a/src/protocol/peer_connection_base.h
+++ b/src/protocol/peer_connection_base.h
@@ -69,8 +69,8 @@ public:
   bool                is_seeder() const               { return m_peerChunks.is_seeder(); }
   bool                is_not_seeder() const           { return !m_peerChunks.is_seeder(); }
 
-  bool                is_encrypted() const            { return m_encryption.is_encrypted(); }
-  bool                is_obfuscated() const           { return m_encryption.is_obfuscated(); }
+  bool                is_encrypted() const            { return m_encryption->is_encrypted(); }
+  bool                is_obfuscated() const           { return m_encryption->is_obfuscated(); }
 
   PeerInfo*           mutable_peer_info()             { return m_peerInfo; }
 
@@ -196,7 +196,7 @@ protected:
   uint32_t            m_extensionOffset;
 
   EncryptBuffer*      m_encryptBuffer{};
-  EncryptionInfo      m_encryption;
+  EncryptionInfo*     m_encryption;
   ProtocolExtension*  m_extensions{};
 
   bool m_incoreContinous{false};
diff --git a/src/protocol/peer_connection_leech.cc b/src/protocol/peer_connection_leech.cc
index ac92c3a95ce0..9d8ea3c30a08 100644
--- a/src/protocol/peer_connection_leech.cc
+++ b/src/protocol/peer_connection_leech.cc
@@ -95,7 +95,7 @@ PeerConnection<type>::receive_keepalive() {
     m_up->write_keepalive();
 
     if (is_encrypted())
-      m_encryption.encrypt(old_end, m_up->buffer()->end() - old_end);
+      m_encryption->encrypt(old_end, m_up->buffer()->end() - old_end);
   }
 
   if (type != Download::CONNECTION_LEECH)
@@ -356,7 +356,7 @@ PeerConnection<type>::event_read() {
           m_down->throttle()->node_used_unthrottled(length);
 
           if (is_encrypted())
-            m_encryption.decrypt(m_down->buffer()->end(), length);
+            m_encryption->decrypt(m_down->buffer()->end(), length);
 
           m_down->buffer()->move_end(length);
         }
@@ -543,7 +543,7 @@ PeerConnection<type>::fill_write_buffer() {
   }
 
   if (is_encrypted())
-    m_encryption.encrypt(old_end, m_up->buffer()->end() - old_end);
+    m_encryption->encrypt(old_end, m_up->buffer()->end() - old_end);
 }
 
 template<Download::ConnectionType type>
diff --git a/src/protocol/peer_connection_metadata.cc b/src/protocol/peer_connection_metadata.cc
index 82851f15f3fb..6bdd6e9b610b 100644
--- a/src/protocol/peer_connection_metadata.cc
+++ b/src/protocol/peer_connection_metadata.cc
@@ -53,7 +53,7 @@ PeerConnectionMetadata::receive_keepalive() {
     m_up->write_keepalive();
 
     if (is_encrypted())
-      m_encryption.encrypt(old_end, m_up->buffer()->end() - old_end);
+      m_encryption->encrypt(old_end, m_up->buffer()->end() - old_end);
   }
 
   return true;
@@ -205,7 +205,7 @@ PeerConnectionMetadata::event_read() {
           m_down->throttle()->node_used_unthrottled(length);
 
           if (is_encrypted())
-            m_encryption.decrypt(m_down->buffer()->end(), length);
+            m_encryption->decrypt(m_down->buffer()->end(), length);
 
           m_down->buffer()->move_end(length);
         }
@@ -291,7 +291,7 @@ PeerConnectionMetadata::fill_write_buffer() {
   }
 
   if (is_encrypted())
-    m_encryption.encrypt(old_end, m_up->buffer()->end() - old_end);
+    m_encryption->encrypt(old_end, m_up->buffer()->end() - old_end);
 }
 
 void
diff --git a/src/utils/rc4.h b/src/utils/rc4.h
index 94fadd8308ab..5b68918ced84 100644
--- a/src/utils/rc4.h
+++ b/src/utils/rc4.h
@@ -47,6 +47,8 @@ class RC4 {
 public:
   RC4();
   ~RC4();
+  RC4(RC4&&) = default;
+  RC4& operator=(RC4&&) = default;
 
   RC4(const unsigned char key[], int len, int enc);
 
```

what do you think?

rakshasa:
Check `encryptionInfo` in the ctor, iirc it is still owned by the caller and can't be passed to PCB like that.

rakshasa:
Add move support to EncryptionInfo, this should simplify the code.

rakshasa:
https://en.cppreference.com/w/cpp/language/move_constructor

Need to read through this to make sure it works correctly.

rakshasa:
It works as if `std::memmove` was called, so it is not sufficient to declare a default move ctor.

You need to define them explicitly and make sure to null out the source object.

neheb:
if only unique_ptr was possible :)

neheb:
moving required too many changes. so copying it is. Hope this is right.

rakshasa:
```
CI results:

failed:    ci-default
failed:    ci-low-diskspace
failed:    ci-nat
failed:    ci-throttle
failed:    ci-torrents
```

Still failing, although the code looks better.

PR #426, opened 2025-05-11
remove some delete calls

neheb:
unique_ptr being a member of a class effectively deletes its copy operations.


rakshasa:
This is bad style imo, you're depending on the behavior of one member variable. (which isn't even the prime reason why we want to disable copying)

PR #427, opened 2025-05-11
clang-tidy: test: use static

neheb:
Found with readability-convert-member-functions-to-static


rakshasa:
Interesting, but it doesn't add anything useful nor does it make the code clearer.

It just adds extra words to the tests.

neheb:
Code gen is improved. But yeah questionable utility for the tests.

PR #430, opened 2025-05-17
use optional

neheb:
Avoids heap allocation.

### DIFF
--- a/src/dht/dht_router.cc
+++ b/src/dht/dht_router.cc
@@ -76,16 +76,13 @@ DhtRouter::DhtRouter(const Object& cache, const rak::socket_address* sa) :
   }
 
   if (m_nodes.size() < num_bootstrap_complete) {
-    m_contacts = std::make_unique<std::deque<contact_t>>();
-
+    m_contacts.emplace();
     if (cache.has_key("contacts")) {
-      const Object::list_type& contacts = cache.get_key_list("contacts");
-
-      for (const auto& contact : contacts) {
-        auto               litr = contact.as_list().begin();
-        const std::string& host = litr->as_string();
-        int port = (++litr)->as_value();
-        m_contacts->emplace_back(host, port);
+      for (const auto& contact : cache.get_key_list("contacts")) {
+        auto litr = contact.as_list().begin();
+        auto host = litr->as_string();
+        auto port = std::next(litr)->as_value();
+        m_contacts->emplace_back(std::move(host), port);
       }
     }
   }
@@ -202,7 +199,7 @@ void
 DhtRouter::add_contact(const std::string& host, int port) {
   // Externally obtained nodes are added to the contact list, but only if
   // we're still bootstrapping. We don't contact external nodes after that.
-  if (m_contacts != NULL) {
+  if (m_contacts.has_value()) {
     if (m_contacts->size() >= num_bootstrap_contacts)
       m_contacts->pop_front();
 
@@ -328,7 +325,7 @@ DhtRouter::store_cache(Object* container) const {
   }
 
   // Insert contacts, if we have any.
-  if (m_contacts != NULL) {
+  if (m_contacts.has_value()) {
     Object& contacts = container->insert_key("contacts", Object::create_list());
 
     for (const auto& m_contact : *m_contacts) {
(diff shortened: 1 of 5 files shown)


src/torrent/peer/client_info.h
```diff
@@ -106,7 +108,7 @@ class LIBTORRENT_EXPORT ClientInfo {
 
   // We don't really care about cleaning up this as deleting an entry
   // form ClientList shouldn't happen.
-  info_type*          m_info;
+  std::optional<info_type> m_info;
```
rakshasa, line 111: Optional has added cost of a bool, and isn't necessary here as we should never have a null value.

In fact we can remove the m_info completely and just store the `const char*`, which all get initialized from static DATA.

neheb: Adjusted.

src/torrent/peer/peer_list.h
```diff
@@ -86,7 +85,7 @@ class LIBTORRENT_EXPORT PeerList : private std::multimap<socket_address_key, Pee
   static ipv4_table   m_ipv4_table;
 
   DownloadInfo*       m_info;
-  std::unique_ptr<AvailableList> m_available_list;
+  AvailableList       m_available_list{};
```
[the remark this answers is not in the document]
neheb: so even though this is the case, It's all a part of libtorrent_torrent. Maybe OK?

src/tracker/tracker_udp.h
```diff
@@ -75,8 +75,8 @@ class TrackerUdp : public SocketDatagram, public TrackerWorker {
   uint64_t            m_connection_id{};
   uint32_t            m_transaction_id{};
 
-  std::unique_ptr<ReadBuffer>  m_read_buffer;
-  std::unique_ptr<WriteBuffer> m_write_buffer;
+  std::optional<ReadBuffer>  m_read_buffer;
+  std::optional<WriteBuffer> m_write_buffer;
```
rakshasa, line 79: There's no need to 're-allocate' the buffers, that was from before a rewrite so they can just be included directly.

src/torrent/peer/client_info.h
```diff
@@ -106,7 +102,7 @@ class LIBTORRENT_EXPORT ClientInfo {
 
   // We don't really care about cleaning up this as deleting an entry
   // form ClientList shouldn't happen.
-  info_type*          m_info;
+  const char* m_info{};
```
rakshasa, line 105: Align properly.

src/tracker/tracker_udp.h
neheb: No idea what you mean so removed.

src/torrent/peer/client_info.h
neheb: Done.

src/torrent/peer/peer_list.h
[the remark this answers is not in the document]
neheb: Removed.

src/torrent/peer/client_info.h
```diff
@@ -86,8 +82,8 @@ class LIBTORRENT_EXPORT ClientInfo {
 protected:
   void                set_type(id_type t)                     { m_type = t; }
   
-  info_type*          info() const                            { return m_info; }
-  void                set_info(info_type* ptr)                { m_info = ptr; }
+  const char* info() const                                    { return m_info; }
+  void                set_info(const char* ptr)               { m_info = ptr; }
```
rakshasa, line 86: You need to fix this align also.

src/dht/dht_router.cc
```diff
@@ -76,16 +76,11 @@ DhtRouter::DhtRouter(const Object& cache, const rak::socket_address* sa) :
...
-      const Object::list_type& contacts = cache.get_key_list("contacts");
-
-      for (const auto& contact : contacts) {
-        auto               litr = contact.as_list().begin();
-        const std::string& host = litr->as_string();
-        int port = (++litr)->as_value();
-        m_contacts->emplace_back(host, port);
+      for (const auto& contact : cache.get_key_list("contacts")) {
+        auto litr = contact.as_list().begin();
+        m_contacts->emplace_back(litr->as_string(), std::next(litr)->as_value());
```
rakshasa, line 83: The host/port variables in the original code serve as comments, and are therefore preferred imo. The code output is the same.

neheb: OK

src/torrent/peer/client_info.h
neheb: Done.

PR #439, opened 2025-05-22
sha1: remove more leaks

neheb:


neheb:
Just to make sure, I reverted to the original SHA_CTX implementation and the leak numbers are the same.

rakshasa:
This makes the life-cycle of this class unsafe, and the reason why you get twice the leaks is the default ctor doing new. Changed.

https://mails.dpdk.org/archives/dev/2023-November/281012.html

Also it might be an issue with the openssl implementation having a leak.

rakshasa:
fix/sha1-deleter

neheb:
huh? the leaks go down.

I don't think that openssl patch is related.

rakshasa:
They go down because you explicitly free the context in final_c.

This means Sha1 objects are being leaked, and your change only reduces how much memory this leak leaks.

neheb:
I'm not seeing that in the Address Sanitizer output.

btw the default constructor does nothing to m_ctx. if default initialized it'll set m_ctx to nullptr but no more than that.

m_ctx = new EVP_MD_CTX doesn't even work as it's opaque.

rakshasa:
Default ctor in the deleter, not ctx.

neheb:
Let's just not waste any more time.

rakshasa:
As said, we need to figure out where (I assume) HashChunk is leaking instead of focusing on down-stream leaks.

neheb:
current master:

```
32869:SUMMARY: AddressSanitizer: 98552 byte(s) leaked in 1013 allocation(s).
33184:SUMMARY: AddressSanitizer: 204200 byte(s) leaked in 3063 allocation(s).
```

After this PR.

```
32869:SUMMARY: AddressSanitizer: 98552 byte(s) leaked in 1013 allocation(s).
32992:SUMMARY: AddressSanitizer: 122520 byte(s) leaked in 1021 allocation(s).

```

neheb:
Sigh.

you can get rid of the _ex suffix in the final function to remove some of the leaks. Even after doing that I don’t remember matching the 1021 number but they did go down.

rakshasa:
We see the same behavior as the bug report I linked, as such it seems a very likely candidate.

I've added error checks to SHA1 that throws internal_error. If correct this will trigger when you run it now.

neheb:
That bug report is completely unrelated.

The code as is is wrong and leaks memory.

neheb:
Since it seems this will never be merged, closing.

PR #442, opened 2025-05-23
Fix/sha1 deleter

rakshasa:


neheb:
yeah the first commit does nothing.

```
37581:SUMMARY: AddressSanitizer: 368600 byte(s) leaked in 3103 allocation(s).
```

before and after it.

rakshasa:
I'd rather get the deleter working as similar code is used other places.

Added default ctor as it seems to be required for default deleter, and tested it gets called.

rakshasa:
It's more important to figure out why this is leaking.

Issue #440, opened 2025-05-22
OpenBSD test fails


neheb:
I was able to get meson running on the CI and found these:

```
2025-05-22T21:40:12.6802296Z test_tracker_timeout::test_timeout_tracker : error
2025-05-22T21:40:12.6802539Z stderr:
2025-05-22T21:40:12.6802893Z assertion "is_inactive()" failed: file "../src/torrent/utils/thread.cc", line 82, function "stop_thread_wait"
```

```
2025-05-22T21:40:02.6268540Z ai_ref_err got wrong error, expected 'name or service is not known', got 'no address associated with name'
2025-05-22T21:40:02.6269325Z 
2025-05-22T21:40:02.6269845Z ai_ref_err got wrong error, expected 'name or service is not known', got 'non-recoverable failure in name resolution'
2025-05-22T21:40:02.6270530Z 
2025-05-22T21:40:02.6271040Z ai_ref_err got wrong error, expected 'no address associated with name', got 'non-recoverable failure in name resolution'
2025-05-22T21:40:02.6271617Z 
2025-05-22T21:40:02.6272206Z ai_ref_err got wrong error, expected 'temporary failure in name resolution', got 'non-recoverable failure in name resolution'
2025-05-22T21:40:02.6272907Z 
2025-05-22T21:40:02.6273527Z ai_ref_err got wrong error, expected 'address family for name not supported', got 'non-recoverable failure in name resolution'
2025-05-22T21:40:02.6274400Z  : assertion
```

That last one is very strange. Probably some vmactions oddity.

I also needed to pass TEST_IGNORE_SIGNAL_INTERRUPT_LATENCY since it fails there as well.

PR #458, opened 2025-05-27
fix getaddrinfo test with OpenBSD

neheb:
It throws EAI_FAIL for some reason.

Get rid of EAI_ADDRFAMILY from IPv4 address. It doesn't throw anywhere.


neheb:
Fixes: https://github.com/rakshasa/libtorrent/issues/440

the thread crash was fixed with the poll commit I think.

PR #471, opened 2025-05-29
Abort on SHA1 errors.

rakshasa:


src/utils/sha1.h
```diff
@@ -29,17 +31,20 @@ Sha1::init() {
...
 Sha1::update(const void* data, unsigned int length) {
-  EVP_DigestUpdate(m_ctx.get(), data, length);
+  if (EVP_DigestUpdate(m_ctx.get(), data, length) == 0)
+    throw internal_error("Sha1::update() failed to update SHA-1 context.");
 }
 
 inline void
 Sha1::final_c(void* buffer) {
-  EVP_DigestFinal_ex(m_ctx.get(), static_cast<unsigned char*>(buffer), nullptr);
+  if (EVP_DigestFinal_ex(m_ctx.get(), static_cast<unsigned char*>(buffer), nullptr) == 0)
```
neheb, line 46: Since you’re dead set on using EVP, remove the _ex suffix here. It calls reset to free some memory

PR #491, opened 2025-06-09
some cleanups

neheb:


src/torrent/net/socket_address.cc
```diff
@@ -136,8 +136,8 @@ sa_unique_ptr
 sa_make_unspec() {
   sa_unique_ptr sa(new sockaddr);
 
-  std::memset(sa.get(), 0, sizeof(sockaddr));
-  sa.get()->sa_family = AF_UNSPEC;
+  *sa = sockaddr{};
```
rakshasa: You can change the above to `sa_unique_ptr sa(new sockaddr{});` to clean it up more.

src/torrent/net/utils.cc
```diff
@@ -68,14 +65,11 @@ auto detect_local_sin6_addr() -> sin6_unique_ptr {
     return sin6_unique_ptr();
   }
 
-  // TODO: Check if unique_ptr works.
-  std::shared_ptr<void> _fd(nullptr, [fd](...){ fd_close(fd); });
```
rakshasa, line 72: This ensures the fd is closed at return of the function.

It's ugly, and should probably be explicit calls to `fd_close` at each of the return paths.

neheb: Oh I didn’t realize the outer fd was captured in the lambda. There’s a way to make unique_ptr handle ints like this directly with some boilerplate.

rakshasa: Just replace it with fd_close on all return paths is cleaner and requires less future readers of the code.

neheb: Hmmm for even more simplicity, GCC and Clang’s cleanup attribute can be used.

rakshasa: Trying to avoid using compiler-specific attributes, especially esoteric ones.

Also, just realized that we can't use unique/shared_ptr deleters as a hack, as those types don't call the deleter when the object is null.

src/torrent/net/socket_address.cc
neheb: Fixed.

src/torrent/net/utils.cc
neheb: Fixed.

rakshasa:
```
#46 34.17 /build/rtorrent/source/src/core/download.cc:132:44: error: member access into incomplete type 'torrent::File'
#46 34.19   132 |        !file_stat.update(file_list->front()->frozen_path()))) {
#46 34.19       |                                            ^
#46 34.19 /prefix/include/torrent/common.h:57:7: note: forward declaration of 'torrent::File'
#46 34.19    57 | class File;
#46 34.19       |       ^
```

neheb:
https://github.com/rakshasa/rtorrent/pull/1517

PR #493, opened 2025-06-11
some unique_ptr cleanups

neheb:


src/torrent/net/resolver.cc
```diff
@@ -44,15 +44,15 @@ Resolver::resolve_preferred(void* requester, const std::string& hostname, int fa
       if (err == 0) {
         if (sin != nullptr && sin6 != nullptr) {
           if (preferred == AF_INET)
-            result = sa_copy_in(sin.get());
+            result = std::reinterpret_pointer_cast<sockaddr>(sin);
```
rakshasa: Copying is required here.

neheb: std::reinterpret_pointer_cast copies, no?

rakshasa: No, it shares the underlying use count.

However my real point is that a copy is intentionally done here as we're passing data from one thread to another, and the source thread does it's own cleanup.

There's no reason to try to optimize this since we want to be as safe as possible.

neheb: removed.

src/download/download_wrapper.h
```diff
@@ -74,7 +74,7 @@ class DownloadWrapper {
   void                finished_download();
 
   std::unique_ptr<DownloadMain> m_main;
-  std::unique_ptr<Object>       m_bencode;
+  Object*                       m_bencode;
```
rakshasa: It seems you're reverting changes.

src/manager.cc
```diff
@@ -43,15 +43,15 @@ void event_remove_and_close(Event* event) { utils::ThreadInternal::poll()->remov
...
-    m_dht_controller(new tracker::DhtController),
+  : m_chunk_manager(std::make_unique<ChunkManager>()),
+    m_connection_manager(std::make_unique<ConnectionManager>()),
+    m_download_manager(std::make_unique<DownloadManager>()),
+    m_file_manager(std::make_unique<FileManager>()),
+    m_handshake_manager(std::make_unique<HandshakeManager>()),
+    m_resource_manager(std::make_unique<ResourceManager>()),
+
+    m_client_list(std::make_unique<ClientList>()),
+    m_dht_controller(std::make_unique<tracker::DhtController>()),
```
rakshasa: These don't actually do anything to clean up the code.

src/download/download_wrapper.h
neheb: This was not done by me: https://github.com/rakshasa/libtorrent/commit/53f7062513676

src/manager.cc
neheb: hmm? what's wrong with matching the constructors with make_unique? It's clearer what the type is.

src/download/download_wrapper.h
rakshasa: You're reverting the change change in that ref afaics.

src/manager.cc
rakshasa: The type is in the class declaration, and we can't mistakenly pass the wrong types to the ctor.

Difference from other cases is that here we are calling the unique_ptr ctor directly, while make_unique is a helper method to call unique_ptr ctor.

neheb: so for consistency, do you recommend replacing make_unique in constructors with new?

rakshasa: Depends, for initializing member variables in the constructor as above, then yes.

However when you pass a unique_ptr to a constructor within a code block, then using make_unique actually tells the reader useful information. (and can catch / make obvious bugs if enforced universally)

As such the former it makes sense to make it less verbose, while the latter you want to write it out by using the _helper function_ make_unique.

neheb: OK done

src/torrent/data/file_list.cc
```diff
@@ -592,7 +592,7 @@ FileList::create_chunk(uint64_t offset, uint32_t length, bool hashing, int prot)
   if (offset + length > m_torrent_size)
     throw internal_error("Tried to access chunk out of range in FileList", data()->hash());
 
-  auto chunk = std::make_unique<Chunk>();
+  auto chunk = new Chunk;
```
rakshasa: This isn't what I meant earlier, this is a use case that make_unique is meant for.

neheb: makes no sense but OK.

rakshasa: It makes no sense because you do not understand my point about why make_unique is superfluous for `m_client_list(std::make_unique<ClientList>())`.

There's no reason why that particular point should mean that the make_unique for chunk above should be removed. It's just copying superficial similarities rather than understanding what is being expressed and the purpose of the make_unique helper function.

rakshasa: Some would say they prefer to use make_unique in both cases, I don't.

This is partially a personal preference based on how I want to express intent in the code.

neheb: I get it for the constructor. Not so much if creating a unique_ptr and releasing it at the end of a function.

rakshasa: That is not a personal preference, if you use raw pointer there you leak when returning nullptr.

src/manager.cc
rakshasa: After considering this, using make_unique would be more consistent with modern C++ and as such probably preferrable.

PR #499, opened 2025-06-13
clang cleanups

neheb:
Fixed warnings found with -Weverything.

### DIFF
--- a/src/net/curl_socket.h
+++ b/src/net/curl_socket.h
@@ -12,7 +12,7 @@ class CurlStack;
 class CurlSocket : public torrent::Event {
 public:
   CurlSocket(int fd, CurlStack* stack) : m_stack(stack) { m_fileDesc = fd; }
-  ~CurlSocket();
+  ~CurlSocket() override;
 
   const char*        type_name() const override { return "curl_socket"; }
 
--- a/src/tracker/tracker_http.h
+++ b/src/tracker/tracker_http.h
@@ -19,7 +19,7 @@ public:
   static constexpr uint32_t http_timeout = 60;
 
   TrackerHttp(const TrackerInfo& info, int flags = 0);
-  ~TrackerHttp();
+  ~TrackerHttp() override;
 
   bool                is_busy() const override;
 


src/data/hash_queue.cc
```diff
@@ -11,8 +11,10 @@
 #include "torrent/exceptions.h"
 #include "torrent/utils/log.h"
 
-#define LT_LOG_DATA(data, log_level, log_fmt, ...)                       \
-  lt_log_print_data(LOG_STORAGE_##log_level, data, "hash_queue", log_fmt, __VA_ARGS__);
+#define LT_LOG_DATA(data, log_level, log_fmt, ...)                                        \
+  do {                                                                                    \
+    lt_log_print_data(LOG_STORAGE_##log_level, data, "hash_queue", log_fmt, __VA_ARGS__); \
+  } while (false)
```
rakshasa: Is this really necessary?

neheb: Alternative is a static void function

neheb: I removed these changes.

rakshasa: Please don't add additional unrelated changes after I've reviewed a PR, this makes it way more complicated than if those changes were to be a separate PR.

neheb: Huh? The changes are the same since posting. I only modified this commit to remove the do while loops

rakshasa: It looked like there was new changes there, however it might just be the original PR was so big I missed them.

In any case, there are too many different kinds of changes in this PR to make it a good merge.

neheb:
rebased. Moved exceptions changes to separate PR.

src/download/delegator.cc
```diff
@@ -22,7 +22,7 @@ Delegator::delegate(PeerChunks* peerChunks, uint32_t affinity, uint32_t maxPiece
   // in progress.
 
   // TODO: What if the hash failed? Don't want data from that peer again.
-  if (affinity >= 0) {
+  if (affinity > 0) {
```
rakshasa: This would be a bug fix, and should be a separate PR to make it easier to track down later.

neheb: split off

PR #502, opened 2025-06-15
avoid unnecessarily copying a weak_ptr

neheb:
Move the function to its own variable.


rakshasa:
This does not change the slots with std::string parameter, so we end up with two arbitrarily different styles.

It's better to be consistent.

rakshasa:
Comparing the code before and after, and the current code just looks way easier to read and not worth reducing such inexpensive copying.

PR #503, opened 2025-06-15
remove = from unsigned check

neheb:
unsigned variables are never negative.


rakshasa:
Have you reviewed the code to check that we actually don't want `0` to be matched?

PR #504, opened 2025-06-15
remove constructor

neheb:
Initialize directly.


neheb:
```
  - Expression: (end_time - start_time) < 2s
```

the solution for macOS is needed on BSD CIs too.

PR #506, opened 2025-06-18
add missing header

neheb:
Needed for AF_INET macro definitions.


neheb:
https://github.com/neheb/rtorrent/actions/runs/15722753880

Not bad.

PR #517, opened 2025-06-21
change failed list to a unique_ptr

neheb:


src/torrent/data/block.h
```diff
@@ -86,8 +86,8 @@ class LIBTORRENT_EXPORT Block {
   void                      change_leader(BlockTransfer* transfer);
   void                      failed_leader();
 
-  BlockFailed*              failed_list()                                { return m_failedList; }
-  void                      set_failed_list(BlockFailed* f)              { m_failedList = f; }
+  const std::unique_ptr<BlockFailed>& failed_list() const;
```
rakshasa: Don't see any benefit of returning unique_ptr here, instead of getting the pointer.

It might have a limited code safety benefit, however it's not enough to counter the added verbosity added to code where e.g. some function is called and `failed_list()` is one of the arguments passed.

neheb:
Rebased.

src/torrent/data/block.h
neheb: Not sure what you mean.

neheb: Changed to use auto here.

rakshasa: Point is if we change these to return std::unique_ptr reference, it should be done everywhere.

And atm I'm leaning towards not, as returning the raw pointer is how it is currently done and I'm not convinced the change is worth it.

neheb: you mean all functions returning .get() ?

rakshasa: Yes, I don't want it returning std::unique_ptr&, but T*.

PR #520, opened 2025-06-27
clang-tidy 21 fixes

neheb:


src/protocol/peer_connection_base.cc
```diff
@@ -323,7 +323,7 @@ PeerConnectionBase::load_up_chunk() {
 
   if (is_encrypted() && m_encryptBuffer == nullptr) {
     m_encryptBuffer = std::make_unique<EncryptBuffer>();
-    m_encryptBuffer->reset();
+    (*m_encryptBuffer).reset();
```
rakshasa: Use `.get()` to make it explicit what we're doing here.

neheb: removed.

PR #523, opened 2025-07-01
fix and simplify atomic check

neheb:
Doesn't really work. Just simplify the whole thing.

### DIFF
--- a/scripts/check_atomic.m4
+++ b/scripts/check_atomic.m4
@@ -30,15 +30,7 @@ AC_DEFUN([CHECK_ATOMIC], [
   AC_LINK_IFELSE([AC_LANG_SOURCE([_CHECK_ATOMIC_testbody])],[
       AC_MSG_RESULT([yes])
     ],[
-      AC_MSG_RESULT([no])
       ATOMIC_LIBS=-latomic
-      AC_MSG_CHECKING([whether std::atomic needs -latomic])
-      AC_LINK_IFELSE([AC_LANG_SOURCE([_CHECK_ATOMIC_testbody])],[
-          AC_MSG_RESULT([yes])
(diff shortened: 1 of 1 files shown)


neheb:
I tested it as not working.

rakshasa:
Does not mean it doesn't (or didn't) work in different environments.

Just disabling the test is not the acceptable solution.

neheb:
Not sure what you mean.

The test basically tests if linking works without libatomic. if it does not, it means we need libatomic. There's no need to test the latter. We know what we need.

After this change it's basically the same thing as the following meson snippet:

```
if cc.links('#include <atomic>\nint main(){atomic_int64_t b;return b.fetch_add(5);}')
  atomic_dep = dependency('', required: false)
else
  atomic_dep = cc.find_library('atomic')
endif
```

Upstream meson's dependency('atomic') sort of does the same thing.

rakshasa:
The removed test checks if adding `-latomic` works, so it's a necessary part of the script.

neheb:
Not really. ATOMIC_LIBS is not passed to the rest so it’s equivalent to the initial. It fails all the same.

rakshasa:
If that was the case, the second tests would fail.

It seems to me ATOMIC_LIBS has a special meaning to the linker.

rakshasa:
If it the second check fails, why aren't you fixing it instead of just removing it?

That's what I can't understand.

neheb:
I think I'll just keep this as a local patch. Maybe someone else wants to deal with this.

neheb:
For reference: https://github.com/openwrt/packages/tree/master/libs/libtorrent/patches

rakshasa:
Test should be fixed now.

PR #524, opened 2025-07-01
replace end() inserts with emplace_back

neheb:
emplace_back returns a reference instead of an iterator. Simpler.

Fixes

```
In file included from ../src/torrent/object.cc:42:
In copy constructor 'torrent::Object::Object(const torrent::Object&)',
    inlined from 'torrent::Object::Object(const torrent::Object&)' at ../src/torrent/object.h:346:1,
    inlined from 'void std::__new_allocator<_Tp>::construct(_Up*, _Args&& ...) [with _Up = torrent::Object; _Args = {torrent::Object}; _Tp = torrent::Object]' at /home/mangix/devstuff/openwrt/staging_dir/toolchain-mips_24kc_gcc-14.3.0_musl/mips-openwrt-linux-musl/include/c++/14.3.0/bits/new_allocator.h:191:4,
    inlined from 'static void std::allocator_traits<std::allocator<_Tp1> >::construct(allocator_type&, _Up*, _Args&& ...) [with _Up = torrent::Object; _Args = {torrent::Object}; _Tp = torrent::Object]' at /home/mangix/devstuff/openwrt/staging_dir/toolchain-mips_24kc_gcc-14.3.0_musl/mips-openwrt-linux-musl/include/c++/14.3.0/bits/alloc_traits.h:575:17,
    inlined from 'std::vector<_Tp, _Alloc>::iterator std::vector<_Tp, _Alloc>::_M_insert_rval(const_iterator, value_type&&) [with _Tp = torrent::Object; _Alloc = std::allocator<torrent::Object>]' at /home/mangix/devstuff/openwrt/staging_dir/toolchain-mips_24kc_gcc-14.3.0_musl/mips-openwrt-linux-musl/include/c++/14.3.0/bits/vector.tcc:371:30,
    inlined from 'std::vector<_Tp, _Alloc>::iterator std::vector<_Tp, _Alloc>::insert(const_iterator, value_type&&) [with _Tp = torrent::Object; _Alloc = std::allocator<torrent::Object>]' at /home/mangix/devstuff/openwrt/staging_dir/toolchain-mips_24kc_gcc-14.3.0_musl/mips-openwrt-linux-musl/include/c++/14.3.0/bits/stl_vector.h:1409:30,
    inlined from 'torrent::Object torrent::object_create_normal(const raw_list&)' at ../src/torrent/object.cc:222:45:
../src/torrent/object.h:355:32: warning: '<unnamed>.torrent::Object::<anonymous>.torrent::Object::<unnamed union>::t_pod' may be used uninitialized [-Wmaybe-uninitialized]
  355 |   case TYPE_VALUE:       t_pod = b.t_pod; break;
      |                          ~~~~~~^~~~~~~~~
../src/torrent/object.cc: In function 'torrent::Object torrent::object_create_normal(const raw_list&)':
../src/torrent/object.cc:222:77: note: '<anonymous>' declared here
  222 |     auto new_entry = result.as_list().insert(result.as_list().end(), Object());
      |                                                                             ^
```

No idea why.

### DIFF
--- a/src/torrent/data/block.cc
+++ b/src/torrent/data/block.cc
@@ -97,18 +97,17 @@ Block::insert(PeerInfo* peerInfo) {
 
   m_notStalled++;
 
-  auto itr = m_queued.insert(m_queued.end(), new BlockTransfer());
+  auto block = new BlockTransfer;
+  block->set_peer_info(peerInfo);
+  block->set_block(this);
+  block->set_piece(m_piece);
+  block->set_state(BlockTransfer::STATE_QUEUED);
+  block->set_request_time(this_thread::cached_seconds().count());
+  block->set_position(0);
+  block->set_stall(0);
+  block->set_failed_index(BlockFailed::invalid_index);
 
-  (*itr)->set_peer_info(peerInfo);
-  (*itr)->set_block(this);
-  (*itr)->set_piece(m_piece);
-  (*itr)->set_state(BlockTransfer::STATE_QUEUED);
-  (*itr)->set_request_time(this_thread::cached_seconds().count());
-  (*itr)->set_position(0);
-  (*itr)->set_stall(0);
-  (*itr)->set_failed_index(BlockFailed::invalid_index);
-
-  return (*itr);
+  return m_queued.emplace_back(block);
 }
   
 void
@@ -193,7 +192,7 @@ Block::transfering(BlockTransfer* transfer) {
     throw internal_error("Block::transfering(...) not queued.");
 
   m_queued.erase(itr);
-  m_transfers.insert(m_transfers.end(), transfer);
+  m_transfers.push_back(transfer);
 
   // If this block already has an active transfer, make this transfer
   // skip the piece. If this transfer gets ahead of the currently
--- a/src/torrent/object.cc
+++ b/src/torrent/object.cc
@@ -218,14 +218,14 @@ Object object_create_normal(const raw_list& obj) {
   raw_list::iterator last = obj.end();
 
   while (first != last) {
-    auto new_entry = result.as_list().insert(result.as_list().end(), Object());
+    auto& new_entry = result.as_list().emplace_back();
 
-    first = object_read_bencode_c(first, last, &*new_entry, 128);
+    first = object_read_bencode_c(first, last, &new_entry, 128);
 
     // The unordered flag is inherited also from list elements who
     // have been marked as unordered, though e.g. unordered strings
     // in the list itself does not cause this flag to be set.
-    if (new_entry->flags() & Object::flag_unordered)
+    if (new_entry.flags() & Object::flag_unordered)
       result.set_internal_flags(Object::flag_unordered);
   }
 
--- a/src/torrent/object_stream.cc
+++ b/src/torrent/object_stream.cc
@@ -113,13 +113,13 @@ object_read_bencode(std::istream* input, Object* object, uint32_t depth) {
 	return;
       }
 
-      auto itr = object->as_list().insert(object->as_list().end(), Object());
-      object_read_bencode(input, &*itr, depth);
+      auto& obj = object->as_list().emplace_back();
+      object_read_bencode(input, &obj, depth);
 
       // The unordered flag is inherited also from list elements who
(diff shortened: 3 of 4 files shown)


src/torrent/data/block.cc
```diff
@@ -97,18 +97,18 @@ Block::insert(PeerInfo* peerInfo) {
...
-  (*itr)->set_block(this);
-  (*itr)->set_piece(m_piece);
-  (*itr)->set_state(BlockTransfer::STATE_QUEUED);
-  (*itr)->set_request_time(this_thread::cached_seconds().count());
-  (*itr)->set_position(0);
-  (*itr)->set_stall(0);
-  (*itr)->set_failed_index(BlockFailed::invalid_index);
-
-  return (*itr);
+  auto& block = m_queued.emplace_back(new BlockTransfer);
```
rakshasa: Write this as `auto*` to be more explicit about code intentions.

src/torrent/utils/log.cc
```diff
@@ -105,7 +105,8 @@ log_rebuild_cache() {
       return entry.equal_outputs(use_outputs);
     });
     if (cache_itr == log_cache.end()) {
-      cache_itr = log_cache.insert(log_cache.end(), log_cache_entry());
+      log_cache.emplace_back();
+      cache_itr = std::prev(log_cache.end());
```
rakshasa: In this case using emplace makes the code less clear.

src/torrent/data/block.cc
neheb: Hmmm using non ref auto is a copy. I guess that’s fine for normal pointers.

rakshasa: It needs to be `auto*`, we're telling the _human reader_ what the code is doing.

neheb: so this is interesting:

```
git grep auto\\\*
src/protocol/peer_chunks.h:  auto*               download_cache()              { return &m_downloadCache; }
src/protocol/peer_chunks.h:  auto*               upload_queue()                { return &m_uploadQueue; }
src/protocol/peer_chunks.h:  const auto*         upload_queue() const          { return &m_uploadQueue; }
src/protocol/peer_chunks.h:  auto*               cancel_queue()                { return &m_cancelQueue; }
```

it's more common to use non pointer auto it seems.
```

File: src/torrent/utils/log.cc (hunk 2 above)
Comment (neheb) [reply src/torrent/utils/log.cc]: Fixed.

File: src/torrent/data/block.cc (hunk 1 above)
Comment (rakshasa) [reply src/torrent/data/block.cc]: Those lines above are a perfect example of why you add the `*`, it makes it explicit that it must be a pointer.

Doesn't mean it has to be done everywhere, however those and the change you made are places where the explicit pointer improves readability and code safety.

Comment (neheb) [reply src/torrent/data/block.cc]: OTOH it’s clear this is a pointer from the right hand side. Which is how auto works anyway.

Comment (neheb) [reply src/torrent/data/block.cc]: auto& is more appropriate IMO. Let’s say this gets converted to a smart pointer. & is needed to avoid copies.

Comment (rakshasa) [reply src/torrent/data/block.cc]: If it gets converted to smart pointer, then the compile fails.

You need to think about it in terms of easy readability, not what a reader can deduce about the return type from staring long enough at the right-hand side.

Comment (neheb) [reply src/torrent/data/block.cc]: Made the change.

**Comment by neheb** [comment]
```
@@ -97,8 +97,7 @@ Block::insert(PeerInfo* peerInfo) {
 
   m_notStalled++;
 
-  auto* block = m_queued.emplace_back(new BlockTransfer);
-
+  auto block = new BlockTransfer;
   block->set_peer_info(peerInfo);
   block->set_block(this);
   block->set_piece(m_piece);
@@ -108,7 +107,7 @@ Block::insert(PeerInfo* peerInfo) {
   block->set_stall(0);
   block->set_failed_index(BlockFailed::invalid_index);
 
-  return block;
+  return m_queued.emplace_back(block);
 }
   
 void
```

might be slightly better. Not sure.

rakshasa:
Yeah, it's better style to initialize the block first.

neheb:
Done.

PR #525, opened 2025-07-02
replace lambdas with mem_fn

neheb:
The latter is more flexible.

### DIFF
--- a/src/net/curl_stack.cc
+++ b/src/net/curl_stack.cc
@@ -123,9 +123,7 @@ CurlStack::close_get(const std::shared_ptr<CurlGet>& curl_get) {
       return;
     }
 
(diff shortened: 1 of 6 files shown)


src/net/curl_stack.cc
```diff
@@ -123,9 +123,8 @@ CurlStack::close_get(const std::shared_ptr<CurlGet>& curl_get) {
       return;
     }
 
-    auto itr = std::find_if(base_type::begin(), base_type::end(), [](auto& curl_get) {
-        return !curl_get->is_active();
-      });
+    auto itr = std::find_if(base_type::begin(), base_type::end(),
```
neheb: wait a minute... this is find_if_not

PR #526, opened 2025-07-03
use pointer for TransferList in header

neheb:


neheb:
I avoided doing

```
@@ -222,9 +222,9 @@ Download::peer_list() const {
   return m_ptr->main()->peer_list();
 }
 
-const TransferList*
+const std::unique_ptr<TransferList>&
 Download::transfer_list() const {
-  return m_ptr->main()->delegator()->transfer_list().get();
+  return m_ptr->main()->delegator()->transfer_list();
 }
 
 ConnectionList*
diff --git a/src/torrent/download.h b/src/torrent/download.h
index d72e88d2f75c..d67e5f86927d 100644
--- a/src/torrent/download.h
+++ b/src/torrent/download.h
@@ -71,7 +71,7 @@ public:
   FileList*           file_list() const;
   PeerList*           peer_list();
   const PeerList*     peer_list() const;
-  const TransferList* transfer_list() const;
+  const std::unique_ptr<TransferList>& transfer_list() const;
 
   ConnectionList*       connection_list();
   const ConnectionList* connection_list() const;
```

since get() seems to be preferred here.

src/download/delegator.h
```diff
@@ -39,7 +43,7 @@ class Delegator {
   void               delegate_new_chunks(std::vector<BlockTransfer*> &transfers, uint32_t maxPieces, PeerChunks* pc, bool highPriority);
   Block*             delegate_seeder(PeerChunks* peerChunks);
 
-  TransferList       m_transfers;
+  std::unique_ptr<TransferList> m_transfers;
```
rakshasa: There's no reason why this needs to be a pointer, we don't gain much in terms of compile time and just ends up adding an indirection during runtime.

neheb: Yeah probably. It's done all over the codebase.

OTOH, transfer_list should definitely be a reference.

rakshasa: Some places it makes sense, others less so. This seems less so.

neheb:
Even with deleted copy operations?

rakshasa:
Not everything has deleted copy operations, so better to have it as a pointer.

PR #531, opened 2025-07-05
replace begin/end with front/back

neheb:
A little cleaner.


src/data/hash_queue.cc
```diff
@@ -109,8 +109,7 @@ HashQueue::work() {
   auto lock = std::scoped_lock(m_done_chunks_lock);
 
   while (!m_done_chunks.empty()) {
-    HashChunk* hash_chunk = m_done_chunks.begin()->first;
-    HashString hash_value = m_done_chunks.begin()->second;
+    auto [hash_chunk, hash_value] = *m_done_chunks.begin();
```
neheb: this actually fails with older clang:

```
../src/data/hash_queue.cc:120:46: error: 'hash_chunk' in capture list does not name a variable
    auto itr = std::find_if(begin(), end(), [hash_chunk](auto& node) { return node.get_chunk() == hash_chunk; });
                                             ^
../src/data/hash_queue.cc:120:99: error: reference to local binding 'hash_chunk' declared in enclosing function 'torrent::HashQueue::work'
    auto itr = std::find_if(begin(), end(), [hash_chunk](auto& node) { return node.get_chunk() == hash_chunk; });
                                                                                                  ^
../src/data/hash_queue.cc:114:11: note: 'hash_chunk' declared here
    auto [hash_chunk, hash_value] = *m_done_chunks.begin();
          ^
2 errors generated.
```

that is a hell of a bug.

src/data/chunk_list.cc
```diff
@@ -171,7 +171,7 @@ ChunkList::release(ChunkHandle* handle, release_flags flags) {
   if (!handle->is_valid())
     throw internal_error("ChunkList::release(...) received an invalid handle.");
 
-  if (handle->object() < &*begin() || handle->object() >= &*end())
+  if (handle->object() < &front() || handle->object() >= &back())
```
rakshasa: End is off by one, should be `>`.

However I don't like this change as begin/end are more natural for iterators/position comparisons.

neheb: I assume you mean &front/back and not simple front/back.

rakshasa: No, I mean it is more natural to use begin/end when dealing with comparison of position.

neheb: Removed.

PR #535, opened 2025-07-11
use try_emplace

neheb:
Allows removal of make_pair. It's also a bit clearer.


src/dht/dht_router.cc
```diff
@@ -513,7 +513,7 @@ DhtRouter::split_bucket(const DhtBucketList::iterator& itr, DhtNode* node) {
     throw internal_error("DhtRouter::split_bucket router ID ended up in wrong bucket.");
 
   // Insert new bucket with iterator hint = just before current bucket.
-  auto other = m_routingTable.emplace_hint(itr, newBucket->id_range_end(), newBucket);
+  auto other = m_routingTable.try_emplace(itr, newBucket->id_range_end(), newBucket);
```
rakshasa: These have different behaviors.

neheb: Hmm OK,

PR #536, opened 2025-07-11
add some const

neheb:


src/tracker/tracker_worker.h
```diff
@@ -67,7 +67,7 @@ class TrackerWorker {
 
   bool                is_busy_not_scrape() const;
 
-  std::string         tracker_id() const                    { return m_tracker_id; }
+  const auto&         tracker_id() const                    { return m_tracker_id; }
```
rakshasa: This is incorrect as it requires a copy due to thread safety.

neheb: Removed.

src/net/throttle_node.h
```diff
@@ -28,8 +28,7 @@ class ThrottleNode {
   void                clear_quota()                   { m_quota = 0; }
   void                set_quota(uint32_t q)           { m_quota = q; }
 
-  iterator            list_iterator()                 { return m_listIterator; }
-  const_iterator      list_iterator() const           { return m_listIterator; }
+  iterator            list_iterator() const           { return m_listIterator; }
```
rakshasa: This changes the permissions, e.g. it allows us to get mutable objects from const parent which is contrary to intent.

src/protocol/extensions.h
```diff
@@ -80,7 +80,7 @@ class ProtocolExtension {
   void                read_start(int type, uint32_t length, bool skip);
   bool                read_done();
 
-  char*               read_position()                  { return m_readPos; }
+  char*               read_position() const            { return m_readPos; }
```
rakshasa: Same here, this should not be allowed for const objects.

src/torrent/peer/peer_info.h
```diff
@@ -105,7 +105,7 @@ class LIBTORRENT_EXPORT PeerInfo {
   // Internal to libTorrent:
   //
 
-  PeerConnectionBase* connection()                          { return m_connection; }
+  PeerConnectionBase* connection() const                    { return m_connection; }
```
rakshasa: Same.

neheb: Removed.

src/protocol/extensions.h
neheb: Removed.

src/net/throttle_node.h
neheb: Removed.

rakshasa:
This is too big for no real gain in code quality.

neheb:
Rebased. you want this split into smaller commits?

src/download/download_main.h
```diff
@@ -84,9 +84,9 @@ class DownloadMain {
   group_entry*        up_group_entry()                           { return &m_up_group_entry; }
   group_entry*        down_group_entry()                         { return &m_down_group_entry; }
 
-  DataBuffer          get_ut_pex(bool initial)                   { return (initial ? m_ut_pex_initial : m_ut_pex_delta).clone(); }
+  DataBuffer          get_ut_pex(bool initial) const             { return (initial ? m_ut_pex_initial : m_ut_pex_delta).clone(); }
```
rakshasa, line 87: All of these above should be non-const as they return mutable pointers.

neheb: Fixed I think. DataBuffer here is not a pointer.

neheb:
Rebased.

rakshasa:
Doesn't really improve the code enough to be worth reviewing.

PR #537, opened 2025-07-20
cppcheck fixes

neheb:


src/dht/dht_tracker.cc
```diff
@@ -47,7 +47,7 @@ DhtTracker::get_peers(unsigned int maxPeers) {
     throw internal_error("DhtTracker::BencodeAddress is packed incorrectly.");
 
   auto first = m_peers.begin();
-  auto last  = m_peers.end();
+  auto last  = std::prev(m_peers.end());
```
rakshasa: This looks wrong, we should rather fix how the second argument to raw_list is derived when `last == end()`.

src/download/download_main.cc
```diff
@@ -397,7 +397,7 @@ DownloadMain::do_peer_exchange() {
     if (!message->empty() && (message->data() == m_ut_pex_initial.data() || message->data() == m_ut_pex_delta.data())) {
       auto buffer = new char[message->length()];
       memcpy(buffer, message->data(), message->length());
-      message->set(buffer, buffer + message->length(), true);
+      message->set(buffer, message->length(), true);
```
rakshasa: If you change the behavior, rename it to something like `set_data`.

src/net/data_buffer.h
```diff
@@ -45,24 +45,29 @@ namespace torrent {
...
+  DataBuffer(char* data, size_t end) :
+      m_data(data), m_end(end) {}
 
   DataBuffer          clone() const        { DataBuffer d = *this; d.m_owned = false; return d; }
-  DataBuffer          release()            { DataBuffer d = *this; set(NULL, NULL, false); return d; }
+  DataBuffer          release() {
+    DataBuffer d = *this;
+    set(nullptr, 0, false);
+    return d;
+  }
```
rakshasa: We don't do multiline functions in struct/class, these should be outside.

src/net/data_buffer.h
```diff
@@ -45,24 +45,29 @@ namespace torrent {
...
+  size_t              length() const       { return m_end; }
 
   void                clear();
-  void                set(char* data, char* end, bool owned);
+  void                set(char* data, size_t end, bool owned);
 
 private:
   char*               m_data{};
-  char*               m_end{};
+  size_t              m_end{};
```
rakshasa: This should be renamed.

src/dht/dht_tracker.cc
neheb: I moved this change to the end. Not sure if this is what you meant.

src/download/download_main.cc
neheb: Done.

src/net/data_buffer.h
neheb: Moved them down to the inline ones.

src/net/data_buffer.h
neheb: Done.

src/dht/dht_tracker.cc
rakshasa: No, you need to change how the `bencode()` distance is calculated, without changing `last`.

It should be `std::distance(first, last) * sizeof(BencodeAddress)` afaics.

neheb: OK

src/net/data_buffer.h
```diff
@@ -45,43 +45,58 @@ namespace torrent {
...
+DataBuffer::clone() const {
+  DataBuffer d = *this;
+  d.m_owned = false;
+  return d;
+}
+
+inline DataBuffer
+DataBuffer::release() {
+  set_data(nullptr, 0, false);
+  return *this;
```
rakshasa: This changes behavior.

src/protocol/extensions.cc
```diff
@@ -409,7 +409,7 @@ ProtocolExtension::send_metadata_piece(size_t piece) {
   m_pending = build_bencode((2 * sizeof(size_t)) + length + 120, "d8:msg_typei1e5:piecei%zue10:total_sizei%zuee", piece, metadataSize);
 
   memcpy(m_pending.end(), buffer + (piece << metadata_piece_shift), length);
-  m_pending.set(m_pending.data(), m_pending.end() + length, m_pending.owned());
+  m_pending.set_data(m_pending.data(), length, m_pending.owned());
```
rakshasa: This is incorrect lenght.

neheb: Removed.

src/torrent/error.cc
```diff
@@ -77,7 +77,7 @@ static constexpr std::array<const char*, e_last + 1> errorStrings{
 
 const char*
 strerror(int err) {
-  if (err < 0 || err > e_last)
+  if (err < 0 || err >= e_last)
```
rakshasa: This is incorrect, we use `e_last+1` above.

neheb: removed.

rakshasa:
Not needed.

PR #538, opened 2025-07-29
clang: fix more weak vtables warnings

neheb:
Also move a bunch of static functions to anonymous namespaces.

basically ones found with

```
rg Wweak v | grep -v exceptions\\\.h  | grep -v test/
```

test/ probably doesn't matter and exceptions.h is...interesting. Can't tell if the comment in exceptions.cc is relevant:

```
// Use actual functions, instead of inlined, for the ctor of
// exceptions. This allows us to create breakpoints at throws. This is
// limited to rarely thrown exceptions.
```

### DIFF
--- a/src/dht/dht_server.cc
+++ b/src/dht/dht_server.cc
@@ -22,6 +22,23 @@
 #define LT_LOG_THIS(log_fmt, ...)                                       \
   lt_log_print_subsystem(torrent::LOG_DHT_SERVER, "dht_server", log_fmt, __VA_ARGS__);
 
+namespace {
+// Error in DHT protocol, avoids std::string ctor from communication_error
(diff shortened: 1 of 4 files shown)


src/torrent/utils/uri_parser.h
```diff
@@ -37,11 +37,6 @@ struct uri_query_state : public uri_base_state {
   uri_query_list elements;
 };
 
-class uri_error : public ::torrent::input_error {
-public:
-  using input_error::input_error;
-};
-
```
rakshasa: This needs to be in the header as we want to be able to catch errors by type.

neheb: Removed.

neheb:
not sure what this is a reference to.

PR #550, opened 2025-08-31
DhtTransaction: Hash the IPv6 addresses

jmakovicka:


src/dht/dht_transaction.cc
```diff
@@ -251,18 +252,26 @@ DhtTransaction::~DhtTransaction() {
...
 DhtTransaction::key(const sockaddr* sa, int id) {
-  if (!sa_is_inet(sa))
-    throw internal_error("DhtTransaction::key called with non-inet address.");
-
-  return (static_cast<uint64_t>(reinterpret_cast<const sockaddr_in*>(sa)->sin_addr.s_addr) << 32) + id;
+  key_type result = {};
+  result[16] = id >> 24;
+  result[17] = id >> 16;
+  result[18] = id >> 8;
+  result[19] = id;
```
rakshasa, line 259: This should use htonl.

src/dht/dht_transaction.h
```diff
@@ -252,7 +252,7 @@ class DhtTransaction {
   };
 
   // Key to uniquely identify a transaction with given per-node transaction id.
-  using key_type = uint64_t;
+  using key_type = std::array<uint8_t, 20>;
```
rakshasa, line 255: This misses the point of a key, as in it should be a built-in type that is fast to look up.

We don't care if part of the key gives us the whole ip address, we just want it to be unique.

We could make the most significant bit of the ID part indicate ipv4/6, and have collision check for ipv6 addresses.

jmakovicka: Thanks for the feedback, in a previous iteration, I just xorred the IPv6 address round-robin in the first 7 bytes of the key, which should not collide in practice (especially with the randomized id). I did not run into any trouble with either version of the patch.
https://github.com/rakshasa/libtorrent/commit/7639df0c9847f1dc1feff4e8cd2bb87bd4cc519e

PR #553, opened 2025-09-03
Update configure.ac

beadon:
version in configure is not the version of this code.

This latest code is tagged with a previous version.  updating.


rakshasa:
Master is not the same as stable-0.15, so updating the version would not be correct.

As soon as I make an unstable release I'll update.

PR #580, opened 2025-10-17
Allow exceptions in ~Block and ~ResourceManager by marking noexcept(false)

Auska:
Allow exceptions in ~Block and ~ResourceManager by marking noexcept(false)

### DIFF
--- a/src/torrent/data/block.cc
+++ b/src/torrent/data/block.cc
@@ -38,6 +38,7 @@
 
 #include <algorithm>
 #include <functional>
+#include <cassert>
 
 #include "peer/peer_info.h"
 #include "protocol/peer_connection_base.h"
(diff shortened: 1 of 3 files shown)


Auska:
> The exceptions should be changed to `assert(foo && "text");.

Done.

src/torrent/data/block.cc
```diff
@@ -51,12 +52,10 @@
 namespace torrent {
 
 Block::~Block() {
-  if (m_state != STATE_INCOMPLETE && m_state != STATE_COMPLETED)
-    throw internal_error("Block dtor with 'm_state != STATE_INCOMPLETE && m_state != STATE_COMPLETED'");
+  assert((m_state != STATE_INCOMPLETE && m_state != STATE_COMPLETED) && "Block dtor with 'm_state != STATE_INCOMPLETE && m_state != STATE_COMPLETED'");
```
rakshasa: Assert inverts the check, also make those two separate checks.

Auska: Understood.

src/torrent/data/block.cc
```diff
@@ -74,8 +74,7 @@ Block::~Block() {
   }
   m_transfers.clear();
 
-  if (m_notStalled != 0)
-    throw internal_error("Block::clear() m_stalled != 0.");
+  assert(m_notStalled != 0 && "Block::clear() m_stalled != 0.");
```
rakshasa: As I said earlier, you are inverting these tests (and comments).

Assert fails if false, so this would be `m_notStalled == 0`, we are asserting that `m_notStalled` is zero.

Auska:
I'm not very familiar with assert, which led to this issue.

anthonyryan1:
I think this was superseded by the merged #632 and can probably be closed.

PR #621, opened 2025-12-14
Clear CURL poll event mask on close

jmakovicka:
Avoids an internal error

Poll::process() event_error called but event mask not cleared: name:curl_socket fd:...


rakshasa:
```
void
CurlSocket::close() {
  if (m_fileDesc == -1)
    throw internal_error("CurlSocket::close() m_fileDesc == -1.");

  this_thread::poll()->remove_and_close(this);
```

It is already done.

PR #632, opened 2026-01-14
Fix some compile warnings

fffe:
This is low-priority, I just wanted to clean up some errors in my build.

I think the  `next_event > buffer + 2048` compare is a bug, though one that's probably never going to be hit. I think that should be checking if next_event points within the data returned by read(), not just if it's inside the buffer.

- limit bounds check to data returned by read() in directory_events::event_read()
- add a cast to sizeof() to silence a signed-vs-unsigned comparison warning in directory_events::event_read()
- wrap the LT_LOG_DEBUG_ #defines in an #if 0 to silence multi-line comment warning
- strncpy is being truncated in ClientList::insert(), change it to memcpy to match insert_helper()


fffe:
Yeah, that works.

libtorrent now builds without any errors, at least on my test machine.

PR #635, opened 2026-01-18
UDP tracker DNS thundering herd fix

anthonyryan1:
The change from synchronous to asynchronous DNS with the UDNS library was largely positive.

Prior to that, if a person has tens of thousands of torrents with the same UDP tracker, everything would be stalled waiting for DNS, torrent by torrent.

The TrackerUdp class made a deliberate effort to cache these values, but with TrackerUdp instances existing on a per-torrent basis, the cache was per-torrent as well.

This change replaces the per-instance DNS cache with a static DNS cache (we're still not making any effort to respect TTL at this time, but that behaivour matches the existing and isn't a regression).

Additionally, since this thundering herd problem means tens of thousands of torrents could also request DNS in the amount of time it takes to get the first DNS response, we are allowing TrackerUdp objects to register themselves for a callback when the centralized cache receives a response.

This change goes for a minimally invasive strategy here, and while I've ensured this patch was thoroughly tested, the acceleration here leads to another thundering herd problem further down the stack: Now that DNS is is deduplicated and much faster, TrackerUdp will send tens of thousands of UDP announces and scrapes in a matter of seconds, and risk exhausting file destriptor limits.

That new problem, further down the chain will need attention in a future PR, but I wanted to take a pause and discuss the theme of these bugs.

From an achitectural perspective, seeing the second issue of the same type, unearthed by this improvement: I question if there would be merit to making TrackerUdp instances exist on a per-tracker basis, rather than a per-torrent basis.

The likely next PR in this series about de-duplicating unnecessary work, would be to consolidate UDP fd's on a per-tracker basis. That should also ensure that we're not DoS-ing trackers by sending 50,000 announces in parallel, but instead they would all arrive in a sequence to the tracker and help avoid overloading DoS-ing trackers by throttling the announces to "as fast as they process them from us".

The more state we want to share within TrackerUdp, the less it feels like it should be instanced on a per-torrent basis, and perhaps instead we should have an unordered_map of TrackerUdp objects that all torrents then share.

Thanks for your maintenance of rtorrent, and I look forward to feedback or criticisms of this change.

### DIFF
--- a/src/tracker/tracker_udp.cc
+++ b/src/tracker/tracker_udp.cc
@@ -2,6 +2,7 @@
 
 #include "tracker_udp.h"
 
+#include <algorithm>
 #include <cstdio>
 #include <netdb.h>
 
@@ -24,6 +25,9 @@
                          "tracker_udp", "%p : " log_fmt, static_cast<TrackerWorker*>(this), __VA_ARGS__);
 
 namespace torrent {
+
+// Static shared DNS cache
+TrackerUdp::dns_cache_type TrackerUdp::s_dns_cache;
 
 TrackerUdp::TrackerUdp(const TrackerInfo& info, int flags) :
   TrackerWorker(info, flags) {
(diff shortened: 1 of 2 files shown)


rakshasa:
That's too much code unrelated to UDP tracker requests in `TrackerUDP`.

It should be a separate dns cache class in `ThreadNet`, which is called instead of `ThreadNet::udns()` when cached lookups are wanted.

Then add cached versions of `resolve_*` methods to `torrent::net::Resolver`.

rakshasa:
This is now an issue for some, as reported at the end of https://github.com/rakshasa/rtorrent/issues/1689

Are you working on implementing this PR in the way I suggested above?

PR #650, opened 2026-02-14
Fix use-after-free in DhtSearch

anthonyryan1:
This is a two patch series fixing infrequent crashes in DhtSearch

The first crash and patch patch I successfully traced to precisely what was happening in the code, documented in the commit message.

The second crash, I wasn't able to identify the code path to blame, but I did confirm that converting to shared_ptr (which has some precedence in the codebase) successfully resolves it. I don't think further time spent on it is getting me any closer to finding the lines to blame.

### DIFF
--- a/src/dht/dht_transaction.h
+++ b/src/dht/dht_transaction.h
@@ -51,7 +51,8 @@ struct dht_compare_closer {
 // and returns what nodes to contact with up to three concurrent transactions pending.
 // The map element is the DhtSearch object itself to allow the returned accessors
 // to know which search a given node belongs to.
-class DhtSearch : protected std::map<std::unique_ptr<DhtNode>, DhtSearch*, dht_compare_closer> {
+class DhtSearch : public std::enable_shared_from_this<DhtSearch>,
+                  protected std::map<std::unique_ptr<DhtNode>, DhtSearch*, dht_compare_closer> {
   friend class DhtTransactionSearch;
 
 public:
@@ -304,7 +305,7 @@ public:
   bool                       is_search() override         { return true; }
 
   DhtSearch::const_accessor  node()                       { return m_node; }
-  DhtSearch*                 search()                     { return m_search; }
+  DhtSearch*                 search()                     { return m_search.get(); }
 
   void                       set_stalled();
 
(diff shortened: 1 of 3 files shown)


src/dht/dht_transaction.h
```diff
@@ -44,14 +44,15 @@ struct dht_compare_closer {
...
-  const HashString&    m_target;
+  HashString    m_target;
 };
 
 // DhtSearch contains a list of nodes sorted by closeness to the given target,
 // and returns what nodes to contact with up to three concurrent transactions pending.
 // The map element is the DhtSearch object itself to allow the returned accessors
 // to know which search a given node belongs to.
-class DhtSearch : protected std::map<std::unique_ptr<DhtNode>, DhtSearch*, dht_compare_closer> {
+class DhtSearch : public std::enable_shared_from_this<DhtSearch>,
```
rakshasa, line 54: While convenient, this hides what is happening from the reader and as such obfuscates the code. Also there's times when it can be deleted.

However looking at the code it's a mess, though I wasn't at the time so interested in reviewing the code.

rakshasa:
```
// DhtSearch contains a list of nodes sorted by closeness to the given target,
// and returns what nodes to contact with up to three concurrent transactions pending.
// The map element is the DhtSearch object itself to allow the returned accessors
// to know which search a given node belongs to.
class DhtSearch : protected std::map<std::unique_ptr<DhtNode>, DhtSearch*, dht_compare_closer> {
  friend class DhtTransactionSearch;

public:
  using base_type = std::map<std::unique_ptr<DhtNode>, DhtSearch*, dht_compare_closer>;

  // Number of closest potential contact nodes to keep.
  static constexpr unsigned int max_contacts = 18;

  // Number of closest nodes we actually announce to.
  static constexpr unsigned int max_announce = 3;

  DhtSearch(const HashString& target, const DhtBucket& contacts);
  virtual ~DhtSearch();

  // Wrapper for iterators, allowing more convenient access to the key
  // and element values, which also makes it easier to change the container
  // without having to modify much code using iterators.
  template <typename T>
  struct accessor_wrapper : public T {
    accessor_wrapper() = default;
    accessor_wrapper(const T& itr) : T(itr) { }

    const std::unique_ptr<DhtNode>& node() const     { return (**this).first; }
    DhtSearch*                      search() const   { return (**this).second; }
  };
```

The top comment makes it clear this needs to be rewritten. Essentially remove `DhtSearch*` from the map, only have `DhtNode` as shared_ptr.

Change `accessor_wrapper` to be just a normal struct with DhtNode/DhtSearch, both shared ptrs. `get_contact()` returns just the `DhtNode`, and construct `accessor_wrapper` when creating transactions.

rakshasa:
I've converted things to use std::shared_ptr in a branch, and will merge that asap.

PR #691, opened 2026-04-28
Fix 3 different crashes of rtorrent, when build as a fully static binary using MUSL cross compiler.

Trogious:
There are 2 fixes:
1. Handle EBADF and ENOENT gracefully in epoll_ctl MOD/DEL
2. Set explicit 8 MB thread stack size in pthread_create

Detailed explanation of issues, when they happen and how they're fixed, are inside each commit.
Both fixes I've verified to be working - crashes in all 3 cases (EBADF, ENOENT, pthread stack size) aren't happening anymore.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-AC_INIT([[libtorrent]],[[0.16.10]],[[sundell.software@gmail.com]])
+AC_INIT([[libtorrent]],[[0.16.11]],[[sundell.software@gmail.com]])
 
 AC_CONFIG_HEADERS([config.h])
 AC_CONFIG_MACRO_DIRS([scripts])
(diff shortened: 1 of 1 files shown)


Trogious:
Verified on [this binary](https://limewire.com/d/SwFOX#wsBjhPqprI).

rakshasa:
These need to be two different PRs.

---

I'm not sure about straight-forward fixing the stack size to 8MB as it can cause unexpected behavior in other environments.

If there are environments where the small default stack size is causing issues, then it should be fixed for those particular archs.

---

The poll code is supposed to be strict in handling socket states to expose bugs, so the changes as they are aren't really the fix I'd like to merge.

It also leaves SocketManager in a bad state.

rakshasa:
The only real solution is to either have libcurl add a couple changes to their API when it comes to idle connections, and flushing 0-length encrypted data. (which they know about)

Or, what I'm working on, is to separate libcurl and http downloads into a different process using ipc.

https://github.com/rakshasa/shm-ipc

Trogious:
@rakshasa I've applied your feedback in this way:

0. Created 2 separate PRs.
1. the pthread stack size is set to 8MB only if the compiler is MUSL. In this case you would always want to do it. [PR 694](https://github.com/rakshasa/libtorrent/pull/694)
2. epoll descriptor "gymnastics" is now hidden behind a configure flag. [PR 695](https://github.com/rakshasa/libtorrent/pull/695)

PR #695, opened 2026-05-01
Handle EBADF and ENOENT gracefully in epoll_ctl MOD/DEL

Trogious:
Add --enable-epoll-safe-close configure option (disabled by default). When enabled, epoll_ctl errors for MOD and DEL operations are handled gracefully instead of throwing internal_error and aborting the process.

When libcurl manages sockets internally (connection reuse, timeouts, redirects), it may close a file descriptor before libtorrent's poll loop has finished modifying or removing the epoll registration. This race condition manifests as two distinct epoll_ctl errors:

EBADF: curl closed the socket and the fd is now invalid. The kernel has already removed it from all epoll sets automatically.

ENOENT: curl closed the socket and the kernel reassigned the same fd number to a new socket. The new fd is valid but was never added to the epoll set, so EPOLL_CTL_MOD fails because there is nothing to modify.

Previously, both cases threw internal_error which propagated as an uncaught exception in the worker thread, calling std::terminate and aborting the process with messages such as:

  epoll_ctl(MOD) error: name:curl_socket fd:N : Bad file descriptor
  epoll_ctl(MOD) error: name:curl_socket fd:N : in read/write but got ENOENT

The existing ENOENT handling for EPOLL_CTL_MOD attempted a retry via EPOLL_CTL_ADD, but only when the old event mask had no EPOLLIN/EPOLLOUT flags. For curl sockets that were actively reading or writing (the common case), it threw instead of recovering.

With --enable-epoll-safe-close:

  EPOLL_CTL_MOD + EBADF: clear event mask and return (fd is gone,
  kernel already cleaned up).

  EPOLL_CTL_MOD + ENOENT with active read/write mask: clear event mask
  and return instead of throwing (the old socket is gone, curl will
  re-register the new socket through its own callback).

  EPOLL_CTL_MOD + ENOENT without read/write mask: retain the existing
  retry-via-ADD behavior.

  EPOLL_CTL_DEL + EBADF/ENOENT: clear event mask and return (deleting
  an already-gone fd is a no-op).

Without --enable-epoll-safe-close the behavior is unchanged from the current code.

This is consistent with how other event loop implementations (libevent, libev) handle stale file descriptors. This race is more likely to trigger under musl libc but is possible under glibc as well.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,6 +52,14 @@ TORRENT_WITH_SYSROOT
 dnl TORRENT_WITH_XFS
 TORRENT_WITHOUT_KQUEUE
 TORRENT_WITHOUT_EPOLL
+
+AC_ARG_ENABLE(epoll-safe-close,
+  AS_HELP_STRING([--enable-epoll-safe-close],[handle EBADF/ENOENT gracefully in epoll_ctl for externally managed sockets [[default=disable]]]),
+  [
+    if test "$enableval" = "yes"; then
+      AC_DEFINE([EPOLL_SAFE_CLOSE], [1], [Handle EBADF/ENOENT in epoll_ctl MOD/DEL instead of aborting])
+    fi
+  ],[])
 TORRENT_CHECK_FALLOCATE
 TORRENT_WITH_POSIX_FALLOCATE
 TORRENT_WITH_ADDRESS_SPACE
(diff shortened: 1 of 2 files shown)


rakshasa:
Closing this as the errors are supposed to introduce crashes rather than silently failing.

This is related to an ongoing issue with libcurl that needs a different solution.

PR #705, opened 2026-05-08
Avoid scheduling task timeout twice

jmakovicka:
Avoids "called on an already scheduled entry" internal error.

### DIFF
--- a/src/tracker/udp_router.cc
+++ b/src/tracker/udp_router.cc
@@ -496,7 +496,7 @@ UdpRouter::receive_timeout() {
   }
 
   if (!m_timeout_queue.empty())
-    this_thread::scheduler()->wait_until(&m_task_timeout, std::get<1>(m_timeout_queue.front()));
+    this_thread::scheduler()->update_wait_until(&m_task_timeout, std::get<1>(m_timeout_queue.front()));
 }
 
 void


rakshasa:
This should be `update_wait_until()` when the entry might already be scheduled, rather than the check you added.

I used the more strict version because I assumed case this wouldn't happen.

jmakovicka:
Thanks, updated. The stack trace on crash does not look particularly helpful, would need to add some logging to see what is going on. Anyway, with this change, current master looks stable so far for me.
```
#0  torrent::tracker::UdpRouter::receive_timeout (this=0x725c00000380) at tracker/udp_router.cc:499
#1  0x0000555555a7ddf4 in operator() (__closure=0x725c00000498) at tracker/udp_router.cc:32
#2  0x0000555555a83eea in std::__invoke_impl<void, torrent::tracker::UdpRouter::UdpRouter()::<lambda()>&>(std::__invoke_other, struct {...} &) (__f=...) at /usr/include/c++/15/bits/invoke.h:63
#3  0x0000555555a83878 in std::__invoke_r<void, torrent::tracker::UdpRouter::UdpRouter()::<lambda()>&>(struct {...} &) (__fn=...) at /usr/include/c++/15/bits/invoke.h:113
#4  0x0000555555a82d3b in std::_Function_handler<void(), torrent::tracker::UdpRouter::UdpRouter()::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/15/bits/std_function.h:292
#5  0x00005555555c4131 in std::function<void()>::operator() (this=0x725c00000498) at /usr/include/c++/15/bits/std_function.h:593
#6  0x0000555555958b56 in torrent::utils::Scheduler::perform (this=0x720c00000960, current_time=std::chrono::duration = { 1778307980000001us }) at utils/scheduler.cc:171
#7  0x000055555596e46c in torrent::system::Thread::process_events (this=0x726800000600) at system/thread.cc:285
#8  0x000055555596da7f in torrent::system::Thread::event_loop (this=0x726800000600) at system/thread.cc:188
#9  0x0000555555581745 in main (argc=1, argv=0x7fffffffde88) at main.cc:448
```

rakshasa:
Reason why it might already be scheduled is due to `UdpRouter::queue_timeout()` called by `UdpRouter::try_write()`.

So the fix is correct.

PR #714, opened 2026-05-13
Fix use-after-free in dht_compare_closer

jmakovicka:
const HashString& info_hash copy is created in the lambda capture in DhtController::announce

then, the reference to captured value is passed through

DhtRouter::announce
DhtServer::announce
DhtAnnounce::DhtAnnounce
DhtSearch::DhtSearch

The refeenced HashString is then destroyed at the end of callback with the DhtSearch reference outliving it.

Capture the HashString reference instead so DhtSearch still references the value in TrackerWorker::TrackerInfo .

### DIFF
--- a/src/dht/transactions/dht_search.h
+++ b/src/dht/transactions/dht_search.h
@@ -40,7 +40,7 @@ struct dht_compare_closer {
   bool operator () (const std::unique_ptr<DhtNode>& one, const std::unique_ptr<DhtNode>& two) const;
 
 private:
-  const HashString&   m_target;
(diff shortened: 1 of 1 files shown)


rakshasa:
If a download is removed, it deletes the `TrackerInfo`.

Copy it into a shared_ptr instead.

jmakovicka:
Ok, I guess it's simpler to just store the copy in dht_compare_closer then (unless TrackerInfo is refactored to use shared_ptr<HashString>)

PR #724, opened 2026-05-18
socket_manager: add per-category allocation limits

trim21:
Adds per-category tracking within SocketManager so that different socket types can be independently capped via `set_category_max_size()`.

**Categories defined (4):**
- `category_generic` (0) — peer connections, BT accept, uncategorized
- `category_http` (1) — HTTP/curl
- `category_internal` (2) — DHT server, UDP tracker router, thread interrupt pairs, BT listen
- `category_scgi` (3) — SCGI/RPC

**How it works:**
- Each `SocketInfo` stores its category
- Per-category managed counts are tracked alongside the global `m_managed_size`
- When a category's max is set to non-zero, `open_event_or_cleanup()` and `can_open_socket()` additionally check that the category hasn't reached its limit
- `close_event_or_throw()` / `unregister_event_or_throw()` read the stored category from `SocketInfo` to decrement the correct counter

**`set_max_size_and_adjust(max_open)`** is the single entry point for fd allocation:
- Moves all `calculate_*` logic into `socket_manager.cc`
- Computes `m_max_size`, `category_http`, `category_internal` from `max_open`
- Calls `notify_changes()` to notify subscribers

**Subscriber mechanism:**
- `subscribe_to_changes()` / `unsubscribe_from_changes()` follow the same pattern as `NetworkConfig`
- `ThreadNet::init_thread_post_local()` subscribes to update `http_stack` at runtime

**Callers updated to pass explicit categories:**
- `curl_socket.cc` → `category_http`
- `dht_server.cc` → `category_internal`
- `udp_router.cc` → `category_internal`
- `thread.cc` → `category_internal`
- `listen.cc` → `category_internal` (BT listen), `category_generic` (accept)
- `handshake_manager.cc` → `category_generic`
- `download_main.cc` → `category_generic`

This is the libtorrent side of the fix for rakshasa/rtorrent#1777.

### DIFF
--- a/src/dht/dht_server.cc
+++ b/src/dht/dht_server.cc
@@ -5,25 +5,25 @@
 #include <algorithm>
 #include <cstdio>
 
-#include "manager.h"
 #include "dht/dht_bucket.h"
 #include "dht/dht_router.h"
 #include "dht/dht_transaction.h"
 #include "dht/transactions/dht_announce.h"
+#include "manager.h"
 #include "torrent/exceptions.h"
-#include "torrent/object.h"
-#include "torrent/object_static_map.h"
-#include "torrent/object_stream.h"
 #include "torrent/net/fd.h"
 #include "torrent/net/poll.h"
 #include "torrent/net/socket_address.h"
+#include "torrent/object.h"
+#include "torrent/object_static_map.h"
+#include "torrent/object_stream.h"
 #include "torrent/runtime/network_config.h"
 #include "torrent/runtime/runtime.h"
 #include "torrent/runtime/socket_manager.h"
 #include "torrent/utils/log.h"
 #include "tracker/tracker_dht.h"
 
-#define LT_LOG_THIS(log_fmt, ...)                                       \
+#define LT_LOG_THIS(log_fmt, ...) \
   lt_log_print_subsystem(torrent::LOG_DHT_SERVER, "dht_server", log_fmt, __VA_ARGS__);
 
 namespace {
@@ -51,29 +51,31 @@ namespace torrent {
 // See torrent/object_static_map.h for how this works.
 template <>
 const DhtMessage::key_list_type DhtMessage::base_type::keys = {
-  { key_a_id,       "a::id*S" },
-  { key_a_infoHash, "a::info_hash*S" },
-  { key_a_port,     "a::port", },
-  { key_a_target,   "a::target*S" },
-  { key_a_token,    "a::token*S" },
+  {key_a_id, "a::id*S"},
+  {key_a_infoHash, "a::info_hash*S"},
+  {
+    key_a_port,
+    "a::port",
+  },
+  {key_a_target, "a::target*S"},
+  {key_a_token, "a::token*S"},
 
-  { key_e_0,        "e[]*" },
-  { key_e_1,        "e[]*" },
+  {key_e_0, "e[]*"},
+  {key_e_1, "e[]*"},
 
-  { key_q,          "q*S" },
+  {key_q, "q*S"},
 
-  { key_r_id,       "r::id*S" },
-  { key_r_nodes,    "r::nodes*S" },
-  { key_r_token,    "r::token*S" },
-  { key_r_values,   "r::values*L" },
+  {key_r_id, "r::id*S"},
+  {key_r_nodes, "r::nodes*S"},
+  {key_r_token, "r::token*S"},
+  {key_r_values, "r::values*L"},
 
-  { key_t,          "t*S" },
-  { key_v,          "v*" },
-  { key_y,          "y*S" },
+  {key_t, "t*S"},
+  {key_v, "v*"},
+  {key_y, "y*S"},
 };
 
-DhtServer::DhtServer(DhtRouter* router)
-  : m_router(router) {
+DhtServer::DhtServer(DhtRouter* router) : m_router(router) {
 
   m_fileDesc = -1;
   reset_statistics();
@@ -136,8 +138,7 @@ DhtServer::start(int port) {
   runtime::socket_manager()->register_event_or_throw(this, [this]() {
       this_thread::poll()->open(this);
       this_thread::poll()->insert_read(this);
-      this_thread::poll()->insert_error(this);
-    });
+      this_thread::poll()->insert_error(this); }, runtime::SocketManager::category_dht);
 }
 
 void
@@ -154,11 +155,11 @@ DhtServer::stop() {
   this_thread::scheduler()->erase(&m_task_timeout);
 
   runtime::socket_manager()->close_event_or_throw(this, [this]() {
-      this_thread::poll()->remove_and_close(this);
+    this_thread::poll()->remove_and_close(this);
 
-      fd_close(file_descriptor());
-      set_file_descriptor(-1);
-    });
+    fd_close(file_descriptor());
+    set_file_descriptor(-1);
+  });
 
   m_networkUp = false;
 }
@@ -166,10 +167,10 @@ DhtServer::stop() {
 void
 DhtServer::reset_statistics() {
   m_queriesReceived = 0;
-  m_queriesSent = 0;
+  m_queriesSent     = 0;
   m_repliesReceived = 0;
-  m_errorsReceived = 0;
-  m_errorsCaught = 0;
+  m_errorsReceived  = 0;
+  m_errorsCaught    = 0;
 }
 
 // Ping a node whose ID we know.
@@ -198,7 +199,7 @@ DhtServer::find_node(const DhtBucket& contacts, const HashString& target) {
   // This shouldn't happen, it means we had no contactable nodes at all.
   if (!search->start())
     return;
-    // throw internal_error("DhtServer::find_node search start failed, no contactable nodes.");  ///////////////// TMP
+  // throw internal_error("DhtServer::find_node search start failed, no contactable nodes.");  ///////////////// TMP
 
   m_searches.insert(search);
 }
@@ -218,7 +219,7 @@ DhtServer::announce(const DhtBucket& contacts, const HashString& infoHash, std::
   // This can only happen if all nodes we know are bad.
   if (!announce->start())
     return;
-    // throw internal_error("DhtServer::announce search start failed, no contactable nodes.");  ///////////////// TMP
+  // throw internal_error("DhtServer::announce search start failed, no contactable nodes.");  ///////////////// TMP
 
   m_searches.insert(announce);
   announce->update_status();
@@ -343,7 +344,7 @@ DhtServer::create_get_peers_response(const DhtMessage& req, const sockaddr* sa, 
 
   const HashString* info_hash = HashString::cast_from(info_hash_str.data());
 
-  DhtTracker* tracker = m_router->get_tracker(*info_hash, false);
+  DhtTracker*       tracker = m_router->get_tracker(*info_hash, false);
 
   // If we're not tracking or have no peers, send closest nodes.
   if (!tracker || tracker->empty()) {
@@ -380,7 +381,7 @@ DhtServer::create_announce_peer_response(const DhtMessage& req, const sockaddr* 
 void
 DhtServer::process_response(const HashString& id, const sockaddr* sa, const DhtMessage& response) {
   int  transactionId = static_cast<unsigned char>(response[key_t].as_raw_string().data()[0]);
-  auto itr = m_transactions.find(DhtTransaction::key(sa, transactionId));
+  auto itr           = m_transactions.find(DhtTransaction::key(sa, transactionId));
 
   // Response to a transaction we don't have in our table. At this point it's
   // impossible to tell whether it used to be a valid transaction but timed out
@@ -408,17 +409,17 @@ DhtServer::process_response(const HashString& id, const sockaddr* sa, const DhtM
       return;
 
     switch (transaction->type()) {
-      case DhtTransaction::DHT_FIND_NODE:
-        parse_find_node_reply(transaction->as_find_node(), response[key_r_nodes].as_raw_string());
-        break;
+    case DhtTransaction::DHT_FIND_NODE:
+      parse_find_node_reply(transaction->as_find_node(), response[key_r_nodes].as_raw_string());
+      break;
 
-      case DhtTransaction::DHT_GET_PEERS:
-        parse_get_peers_reply(transaction->as_get_peers(), response);
-        break;
+    case DhtTransaction::DHT_GET_PEERS:
+      parse_get_peers_reply(transaction->as_get_peers(), response);
+      break;
 
-      // Nothing to do for DHT_PING and DHT_ANNOUNCE_PEER
-      default:
-        break;
+    // Nothing to do for DHT_PING and DHT_ANNOUNCE_PEER
+    default:
+      break;
     }
 
     // Mark node responsive only if all processing was successful, without errors.
@@ -439,7 +440,7 @@ DhtServer::process_response(const HashString& id, const sockaddr* sa, const DhtM
 void
 DhtServer::process_error(const sockaddr* sa, const DhtMessage& error) {
   int  transactionId = static_cast<unsigned char>(error[key_t].as_raw_string().data()[0]);
-  auto itr = m_transactions.find(DhtTransaction::key(sa, transactionId));
+  auto itr           = m_transactions.find(DhtTransaction::key(sa, transactionId));
 
   if (itr == m_transactions.end())
     return;
@@ -533,26 +534,26 @@ DhtServer::find_node_next(DhtTransactionSearch* transaction) {
 void
 DhtServer::add_packet(std::shared_ptr<DhtTransactionPacket> packet, int priority) {
   switch (priority) {
-    // High priority packets are for important queries, and quite small.
-    // They're added to front of high priority queue and thus will be the
-    // next packets sent.
-    case packet_prio_high:
-      m_highQueue.push_front(std::move(packet));
-      break;
+  // High priority packets are for important queries, and quite small.
+  // They're added to front of high priority queue and thus will be the
+  // next packets sent.
+  case packet_prio_high:
+    m_highQueue.push_front(std::move(packet));
+    break;
 
-    // Low priority query packets are added to the back of the high priority
-    // queue and will be sent when all high priority packets have been transmitted.
-    case packet_prio_low:
-      m_highQueue.push_back(std::move(packet));
-      break;
+  // Low priority query packets are added to the back of the high priority
+  // queue and will be sent when all high priority packets have been transmitted.
+  case packet_prio_low:
+    m_highQueue.push_back(std::move(packet));
+    break;
 
-    // Reply packets will be processed after all of our own packets have been send.
-    case packet_prio_reply:
-      m_lowQueue.push_back(std::move(packet));
-      break;
+  // Reply packets will be processed after all of our own packets have been send.
+  case packet_prio_reply:
+    m_lowQueue.push_back(std::move(packet));
+    break;
 
-    default:
-      throw internal_error("DhtServer::add_packet called with invalid priority.");
+  default:
+    throw internal_error("DhtServer::add_packet called with invalid priority.");
   }
 }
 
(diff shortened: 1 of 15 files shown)


src/dht/dht_server.cc
```diff
@@ -51,29 +51,31 @@ namespace torrent {
...
+  {key_r_values, "r::values*L"},
+
+  {key_t, "t*S"},
+  {key_v, "v*"},
+  {key_y, "y*S"},
 };
 
-DhtServer::DhtServer(DhtRouter* router)
-  : m_router(router) {
+DhtServer::DhtServer(DhtRouter* router) : m_router(router) {
```
rakshasa: Please revert this and the above change, don't modify the coding style of code you don't touch.

Also, the previous style is the preferred one. Same applies to a lot of changes in this PR.

src/torrent/runtime/socket_manager.cc
```diff
@@ -33,12 +33,41 @@ SocketManager::max_size() {
...
+
+void
+SocketManager::account_new_socket(socket_map::iterator itr, uint8_t category) {
+  itr->second.category = category;
+  m_managed_size++;
+  m_category_managed_size[category]++;
+}
+
+void
+SocketManager::account_remove_socket(socket_map::iterator itr) {
```
rakshasa: Add `*_unsafe()` for functions that don't lock.

src/torrent/runtime/socket_manager.cc
```diff
@@ -79,8 +109,11 @@ SocketManager::open_event_or_throw(Event* event, std::function<void ()> func) {
   if (itr != m_socket_map.end()) {
     // We don't allow conflicts for this function.
     LT_LOG("open_event_or_throw() : %s:%s:%i : tried to use an existing file descriptor : %s:%s",
-           this_thread::thread()->name(), event->type_name(), fd,
-           itr->second.thread->name(), itr->second.event->type_name());
+           this_thread::thread()->name(),
+           event->type_name(),
+           fd,
+           itr->second.thread->name(),
+           itr->second.event->type_name());
```
rakshasa: Don't change code like this, it is intentionally formatted as it is.

src/torrent/runtime/socket_manager.cc
```diff
@@ -89,22 +122,26 @@ SocketManager::open_event_or_throw(Event* event, std::function<void ()> func) {
...
 
   LT_LOG("open_event_or_throw() : %s:%s:%i : opened socket",
-         this_thread::thread()->name(), event->type_name(), fd);
+         this_thread::thread()->name(),
+         event->type_name(),
+         fd);
 
   m_socket_map.emplace(fd, SocketInfo{fd, event, this_thread::thread()});
-  m_managed_size++;
+  account_new_socket(m_socket_map.find(fd), category);
```
rakshasa: Don't use find here, we get the iterator for free when inserting.

src/torrent/runtime/socket_manager.h
```diff
@@ -24,28 +25,44 @@ struct SocketInfo {
...
+  // Categories of allocated sockets within the global pool.
+  // When a category's max is non-zero, open operations additionally check
+  // that the category's current usage has not reached its limit.
+  static constexpr uint8_t category_generic  = 0; // peer connections, BT listen, uncategorized
+  static constexpr uint8_t category_http     = 1; // HTTP/curl
+  static constexpr uint8_t category_dht      = 2; // DHT server
+  static constexpr uint8_t category_tracker  = 3; // UDP tracker router
+  static constexpr uint8_t category_internal = 4; // thread interrupt pairs, system signals
+  static constexpr uint8_t category_scgi     = 5; // SCGI/RPC
+  static constexpr uint8_t category_count    = 6;
```
rakshasa: This is too many categories.

We only have 2 sockets for UDP, one for DHT, and a few for internal. Make them all internal. Also, BT listen (not accept) is internal.

Probably should also start adding open files (when doing mmap).

---

These should be proper named enums.

src/torrent/runtime/socket_manager.h
```diff
@@ -24,28 +25,44 @@ struct SocketInfo {
...
+  void     set_max_size(uint32_t max_size);
+  void     set_category_max_size(uint8_t category, uint32_t max_size);
 
-  void                add_unmanaged_socket();
-  void                remove_unmanaged_socket();
+  void     add_unmanaged_socket();
+  void     remove_unmanaged_socket();
 
-  bool                can_open_socket();
+  bool     can_open_socket(uint8_t category = category_generic);
```
rakshasa: Don't add a default parameter value.

src/torrent/runtime/socket_manager.h
```diff
@@ -56,47 +73,52 @@ class LIBTORRENT_EXPORT SocketManager {
   // To avoid reuse race conditions, these calls must also add the Event to read/write polling.
 
   // Throw internal_error on conflicts, as the caller isn't expecting conflicts.
-  void                open_event_or_throw(Event* event, std::function<void ()> func);
+  void open_event_or_throw(Event* event, std::function<void()> func, uint8_t category = category_generic);
```
rakshasa: `void open_event_or_throw(Event* event, uint8_t category, std::function<void()> func);`

src/torrent/runtime/socket_manager.cc
trim21: we have https://github.com/rakshasa/libtorrent/blob/master/.clang-format , would you mind tell me the expected clang-format version?

rakshasa: The formatting is close, but not exactly as expected.

```
diff --git a/.clang-format b/.clang-format
index 6a41c804..ebca7417 100644
--- a/.clang-format
+++ b/.clang-format
@@ -1,6 +1,6 @@
 ---
 BasedOnStyle: LLVM
-Standard: c++14
+Standard: c++17
 
 AllowShortFunctionsOnASingleLine: All
 AllowShortLambdasOnASingleLine: All
@@ -10,11 +10,15 @@ BinPackParameters: false
 BreakConstructorInitializers: AfterColon
 BreakStringLiterals: false
 ColumnLimit: 0
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
 ContinuationIndentWidth: 2
+Cpp11BracedListStyle: false
 IndentCaseLabels: false
 IndentWidth: 2
+PackConstructorInitializers: Never
 PenaltyReturnTypeOnItsOwnLine: 130
 PointerAlignment: Left
+SpacesInContainerLiterals: true
 
 AlignEscapedNewlines: Right
 AlignConsecutiveDeclarations:
```

This might fix some of the issues, however I still prefer not to run clang on unchanged code portions.

src/torrent/runtime/socket_manager.cc
```diff
@@ -33,12 +33,47 @@ SocketManager::max_size() {
...
+void
+SocketManager::account_remove_socket_unsafe(socket_map::iterator itr) {
+  m_managed_size--;
+  m_category_managed_size[itr->second.category]--;
+}
+
+void
+SocketManager::account_replace_socket_unsafe(socket_map::iterator itr, category_t new_category) {
+  m_category_managed_size[itr->second.category]--;
+  m_category_managed_size[new_category]++;
```
rakshasa: There should never be any change in category, and so it only adds code complexity.

src/torrent/runtime/socket_manager.cc
```diff
@@ -33,12 +33,47 @@ SocketManager::max_size() {
...
+SocketManager::account_new_socket_unsafe(socket_map::iterator itr, category_t category) {
+  itr->second.category = category;
+  m_managed_size++;
+  m_category_managed_size[category]++;
+}
+
+void
+SocketManager::account_remove_socket_unsafe(socket_map::iterator itr) {
+  m_managed_size--;
+  m_category_managed_size[itr->second.category]--;
```
rakshasa: Add internal_error exceptions if either is 0.

src/torrent/runtime/socket_manager.cc
```diff
@@ -54,13 +89,14 @@ SocketManager::remove_unmanaged_socket() {
 
 // TODO: Add check to open_event_*.
 bool
-SocketManager::can_open_socket() {
+SocketManager::can_open_socket(category_t category) {
   auto guard = lock_guard();
-  return size() < max_size();
+  return size() < max_size() &&
+         (m_category_max_size[category] == 0 || m_category_managed_size[category] < m_category_max_size[category]);
```
rakshasa: This should be restructured to make it easier to read.

```
if (size() >= max_size())
  return false;

if (m_category_max_size[category] != 0)
  return m_category_managed_size[category] < m_category_max_size[category]);

return true;
```

src/torrent/runtime/socket_manager.cc
```diff
@@ -91,18 +127,19 @@ SocketManager::open_event_or_throw(Event* event, std::function<void ()> func) {
   LT_LOG("open_event_or_throw() : %s:%s:%i : opened socket",
          this_thread::thread()->name(), event->type_name(), fd);
 
-  m_socket_map.emplace(fd, SocketInfo{fd, event, this_thread::thread()});
-  m_managed_size++;
+  auto [itr_new, _] = m_socket_map.emplace(fd, SocketInfo{fd, event, this_thread::thread()});
+  account_new_socket_unsafe(itr_new, category);
```
rakshasa: A little trick:

```
auto itr = m_socket_map.lower_bound(fd);

if (itr != m_socket_map.end() && itr->first == id) {
  // collision
  return;
}

auto itr_new = m_socket_map.emplace_hint(itr, fd, SocketInfo{fd, event, this_thread::thread()});
```

src/torrent/runtime/socket_manager.cc
```diff
@@ -91,18 +127,19 @@ SocketManager::open_event_or_throw(Event* event, std::function<void ()> func) {
...
-SocketManager::open_event_or_cleanup(Event* event, std::function<void ()> func, std::function<void (bool)> cleanup) {
+SocketManager::open_event_or_cleanup(Event* event, category_t category, std::function<void ()> func, std::function<void (bool)> cleanup) {
   auto guard = lock_guard();
 
   if (event->is_open())
     throw internal_error("SocketManager::open_event_or_cleanup(): event is already open");
 
-  if (m_managed_size + m_unmanaged_size >= m_max_size) {
+  if (m_managed_size + m_unmanaged_size >= m_max_size ||
+      (m_category_max_size[category] > 0 && m_category_managed_size[category] >= m_category_max_size[category])) {
```
rakshasa: Isn't this the same as `can_open_socket()`? Add an unsafe version and inline the public version in the header.

src/torrent/runtime/socket_manager.h
```diff
@@ -88,12 +105,19 @@ class LIBTORRENT_EXPORT SocketManager {
...
+  void                account_replace_socket_unsafe(socket_map::iterator itr, category_t new_category);
+
   bool                handle_reused_socket(socket_map::iterator itr);
 
   std::atomic<uint32_t> m_managed_size{};
   std::atomic<uint32_t> m_unmanaged_size{};
   std::atomic<uint32_t> m_max_size{};
 
+  std::array<std::atomic<uint32_t>, category_count> m_category_managed_size{};
+  std::array<std::atomic<uint32_t>, category_count> m_category_max_size{};
```
rakshasa: `using category_list = std::array<std::atomic<uint32_t>, category_count>;`

rakshasa:
The categories are not being set yet. During `torrent::initialize()` the values are set based on the max open sockets value.

These need to be set through SocketManager instead. (e.g. `SocketManager::set_max_size_and_adjust()`)

`torrent::initialize()` then passes the new values when initializing.

After `set_max_size_and_adjust()` is called, registers handler for new updates similar to `NetworkConfig::subscribe_to_changes()`. (do it after the initial call to avoid triggering now)

rakshasa:
Also open file fd's need to be a separate category.

However doing the subscribe_to_changes can be a different PR if it becomes too much work.

trim21:
I

> The categories are not being set yet. During `torrent::initialize()` the values are set based on the max open sockets value.
> 
> These need to be set through SocketManager instead. (e.g. `SocketManager::set_max_size_and_adjust()`)
> 
> `torrent::initialize()` then passes the new values when initializing.
> 
> After `set_max_size_and_adjust()` is called, registers handler for new updates similar to `NetworkConfig::subscribe_to_changes()`. (do it after the initial call to avoid triggering now)

I'm not sure how should I handle this, should I move the `calculate_internal` and `calculate_reserved` to src/torrent/runtime/socket_manager.cc and update all limit when caller set a new max_open (eg, call the `set_max_size_and_adjust`) ?

or we need a new `torrent::set_max_size_and_adjust()` to re-compute all these, including `max_http_connections` when user call `network.max_open_sockets.set`?

rakshasa:
Remove the old `SM::set_max_size()` as the name doesn't fit and the functionality isn't needed. only have the `SM::set_max_size_and_adjust()`.

We should always adjust with how it will be used now.

`network.max_open_sockets.set` should call `SM::set_max_size_and_adjust()` directly and any old code removed.

`calculate_*` should be added to `socket_manager.cc`, and the values they provide are afaics quite decent, however it's been a long time since i checked them.

When adding subscriptions they should be added after initial values are set manually as they will use thread callbacks. (Best place to do it is in init_thread()).

src/torrent/runtime/socket_manager.h
```diff
@@ -86,14 +103,23 @@ class LIBTORRENT_EXPORT SocketManager {
...
   bool                handle_reused_socket(socket_map::iterator itr);
 
   std::atomic<uint32_t> m_managed_size{};
   std::atomic<uint32_t> m_unmanaged_size{};
   std::atomic<uint32_t> m_max_size{};
 
+  category_list       m_category_managed_size{};
+  category_list       m_category_max_size{};
+
   std::mutex          m_mutex;
```
rakshasa:
```
  alignas(std::hardware_destructive_interference_size) std::mutex m_mutex;
```

Something I realized needs to be fixed.

rakshasa: Might end up adding `align_cacheline` to `torrent/common.h` as that's too long.

trim21:
> Remove the old `SM::set_max_size()` as the name doesn't fit and the functionality isn't needed. only have the `SM::set_max_size_and_adjust()`.
> 
> We should always adjust with how it will be used now.
> 
> `network.max_open_sockets.set` should call `SM::set_max_size_and_adjust()` directly and any old code removed.
> 
> `calculate_*` should be added to `socket_manager.cc`, and the values they provide are afaics quite decent, however it's been a long time since i checked them.
> 
> When adding subscriptions they should be added after initial values are set manually as they will use thread callbacks. (Best place to do it is in init_thread()).

how about this?

https://github.com/rakshasa/libtorrent/blob/c604c67346e0ff4c94e66b994184cbac5373252a/src/torrent/torrent.cc#L145-L146

also this: https://github.com/rakshasa/libtorrent/blob/c604c67346e0ff4c94e66b994184cbac5373252a/src/torrent/torrent.cc#L143


something like this? 

image

rakshasa:
No, that's a bad dependency graph.

Point of adding a subscriber to SM is that the code for handling the change is located where the user of the event is.

So net_thread is where http requests are handled, and `src/net/thread_net.cc` is where it gets initialized/shutdown.

The http max size changing needs to be handled in net_thread's thread context, and as such adding the event subscription in `ThreadNet::init_thread()` is a very clean way of doing it.

See `ThreadTracker::init_thread_post_local()` for an example of how to subscribe.

In `torrent::initialize()` you would use `SM::category_size(foo)` to set total and to calculate host connections. The subscription event would also use the same function to calculate host connections.

src/net/thread_net.cc
```diff
@@ -72,10 +88,22 @@ ThreadNet::init_thread() {
 void
 ThreadNet::init_thread_post_local() {
   m_dns_resolver->initialize(this);
+
+  runtime::socket_manager()->subscribe_to_changes(this, [this]() {
+      cancel_callback(this);
+
+      callback(this, [this]() {
+          auto total = runtime::socket_manager()->category_max_size(runtime::SocketManager::category_http);
+          m_http_stack->set_max_host_connections(calculate_max_http_host_connections(total));
```
rakshasa: This should be based on max size.

src/net/thread_net.cc
```diff
@@ -72,10 +88,22 @@ ThreadNet::init_thread() {
...
+          m_http_stack->set_max_total_connections(total);
+        });
+    });
 }
 
 void
 ThreadNet::cleanup_thread() {
+  runtime::socket_manager()->unsubscribe_from_changes(this);
```
rakshasa, line 101: Merge master, callbacks have changed.

src/torrent/runtime/socket_manager.cc
```diff
@@ -13,6 +14,68 @@
...
+  if (open_max >= 16384)
+    return 32;
+  else if (open_max >= 1024)
+    return 16;
+  else
+    return 8;
+}
+
+uint32_t
+calculate_max_http_total_connections(uint32_t open_max) {
```
rakshasa: Name them after the category only, e.g. calculate_http.

src/torrent/runtime/socket_manager.cc
```diff
@@ -44,9 +107,17 @@ SocketManager::category_max_size(category_t category) const {
...
+SocketManager::set_max_size_and_adjust(uint32_t max_open) {
   auto guard = lock_guard();
-  m_max_size = max_size;
+
+  auto max_files = calculate_max_open_files(max_open);
+  auto reserved  = calculate_reserved(max_open);
+  m_max_size = max_open - max_files - reserved;
+
+  m_category_max_size[category_internal] = calculate_internal(max_open);
+  m_category_max_size[category_http]     = calculate_max_http_total_connections(max_open);
```
rakshasa: This should be setting `max_size` to max size.

Get `calculate_reserved`, then set `m_category_max_size` to each of `calculate_http/internal/scgi`.

Last set `m_category_max_size[generic] = max_size - reserved_plus_the_above`.

src/torrent/torrent.cc
```diff
@@ -53,58 +53,17 @@ calculate_max_open_files(uint32_t open_max) {
...
-  if (open_max >= 16384)
+calculate_max_http_host_connections(uint32_t http_total) {
+  if (http_total >= 128)
     return 3;
-  else if (open_max >= 8096)
+  else if (http_total >= 64)
     return 2;
   else // Assumes we don't try less than 64.
     return 1;
 }
```
rakshasa: Remove this function, ThreadNet should have a set_max_total_connections function you can call to reuse the code there.

src/torrent/torrent.cc
```diff
@@ -106,12 +95,9 @@ initialize() {
   auto max_open = this_thread::poll()->open_max();
 
   runtime::socket_manager()->set_max_size_and_adjust(max_open);
-
-  auto http_total = runtime::socket_manager()->category_max_size(runtime::SocketManager::category_http);
-
   manager->file_manager()->set_max_open_files(calculate_max_open_files(max_open));
```
rakshasa: Files are also sockets, and the category needs to be added to SocketManager.

However you don't need to actually register the fds in this PR (with a new register_event_or_cleanup), just at make the category so they get reserved.

src/torrent/runtime/socket_manager.cc
```diff
@@ -110,12 +108,17 @@ void
...
   m_category_max_size[category_internal] = calculate_internal(max_open);
-  m_category_max_size[category_http]     = calculate_max_http_total_connections(max_open);
+  m_category_max_size[category_http]     = calculate_http(max_open);
+  m_category_max_size[category_scgi]     = calculate_scgi(max_open);
+
+  m_category_max_size[category_generic]  = max_open
+    - calculate_reserved(max_open)
+    - m_category_max_size[category_internal]
+    - m_category_max_size[category_http]
+    - m_category_max_size[category_scgi];
```
rakshasa: You should build up the `reserved` value, e.g. do something like:

```
if (max_open < 512)
  throw input_error("foo....");

uint32_t total_allocated{};

auto fn = [this, max_open, &total_allocated](category_t category, uint32_t allocation) {
    total_allocated += allocation;

    if (total_allocated + 8 > max_open)
      throw internal_error("bar....");

     m_category_max_size[category] = allocation;
  });
```

src/net/thread_net.h
```diff
@@ -55,6 +57,8 @@ class LIBTORRENT_EXPORT ThreadNet : public system::Thread {
   std::unique_ptr<net::DnsBuffer>    m_dns_buffer;
   std::unique_ptr<net::DnsCache>     m_dns_cache;
   std::unique_ptr<net::UdnsResolver> m_dns_resolver;
+
+  system::callback_id m_http_subscriber_id;
```
rakshasa: Canonical name is `m_*_callback_id` and they can be used in any number of callbacks, so `m_events_callback_id` makes more sense as it can be used for all events where ThreadTracker is a passive listener.

src/net/thread_net.cc
```diff
@@ -106,6 +104,16 @@ ThreadNet::cleanup_thread() {
   m_dns_resolver->cleanup();
 }
 
+void
+ThreadNet::set_max_connections() {
+  auto sm    = runtime::socket_manager();
+  auto total = sm->category_max_size(runtime::SocketManager::category_http);
+  auto host  = calculate_http_host_connections(sm->max_size());
```
rakshasa: These variable names are too short and makes one spend moments guessing what they are. Just call them `total_size` and `host_size`.

Also no need for `sm`, we want to avoid unnecessary variables like this when it doesn't improve speed nor readability.

src/torrent/runtime/socket_manager.cc
```diff
@@ -108,17 +108,27 @@ void
...
+  uint32_t total_allocated{};
 
-  m_category_max_size[category_internal] = calculate_internal(max_open);
-  m_category_max_size[category_http]     = calculate_http(max_open);
-  m_category_max_size[category_scgi]     = calculate_scgi(max_open);
+  auto set_category = [this, max_open, &total_allocated](category_t category, uint32_t allocation) {
+    total_allocated += allocation;
 
-  m_category_max_size[category_generic]  = max_open
-    - calculate_reserved(max_open)
```
rakshasa: `calculate_reserved` is an extra buffer that must always be there for unexpected fds.

We can't really track absolutely everything, so this margin of reserve is added.

src/torrent/torrent.cc
```diff
@@ -95,7 +79,7 @@ initialize() {
   auto max_open = this_thread::poll()->open_max();
 
   runtime::socket_manager()->set_max_size_and_adjust(max_open);
-  manager->file_manager()->set_max_open_files(calculate_max_open_files(max_open));
+  manager->file_manager()->set_max_open_files(runtime::socket_manager()->category_max_size(runtime::SocketManager::category_files));
```
rakshasa, line 82: File manager lives in ThreadMain, use the same `ThreadMain::m_event_callback_id` to subscribe to SM changes.

trim21:
the ci failed because it doesn't have a compiler support cxx20

```
  checking whether the g++ linker (/usr/bin/ld) supports shared libraries... yes
  checking dynamic linker characteristics... NetBSD ld.elf_so
  checking how to hardcode library paths into programs... immediate
  checking whether g++ supports C++20 features with -std=c++20... no
  checking whether g++ supports C++20 features with +std=c++20... no
  checking whether g++ supports C++20 features with -h std=c++20... no
  checking whether g++ supports C++20 features with -std:c++20... no
  configure: error: *** A compiler with support for C++20 language features is required.
Error: The process '/usr/bin/ssh' failed with exit code 1
```

rakshasa:
Could you force-push this PR into a single commit with a proper (short) commit message?

rakshasa:
Same for the rtorrent PR.

trim21:
can you merge it with squash? That will make it single commit.

rakshasa:
Sure, I do that sometimes for merge conflicts but it changes author so I try to avoid it.

I'll add your tag to the commit message.

trim21:
I mean github web, I just pick a PR and merge it with github web squash and it doesn't change author.

image


And I'm fine with it, feel free to do it.

rakshasa:
Thanks, merging and I'll make a new PR with some minor fixes.

PR #728, opened 2026-05-23
Throw if piece messages are too short

anthonyryan1:


rakshasa:
Should probably be before `if (!m_down->can_read_piece_body())`.

anthonyryan1:
Amended! The NetBSD failure looks like an outage with testing against netbsd rather than the code.

rakshasa:
NetBSD is missing C++20, which likely will be a new requirement due to new callback fixes. (unless there's enough users affected to make the effort for backwards compatibility worth it)

PR #760, opened 2026-06-08
fix: populate empty HeuristicType enum in Download class

trim21:
The `torrent::Download::HeuristicType` enum was declared as `enum HeuristicType {};` with no enumerators. When `option_find_string` returns choke heuristic values (e.g. HEURISTICS_DOWNLOAD_LEECH = 3), the C-style cast to this empty enum type triggers UBSan: 'load of value 3, which is not valid for type HeuristicType'.

Fixed by populating the enum with the four valid choke heuristic values matching `choke_queue::heuristics_enum`.

### DIFF
--- a/src/torrent/download.h
+++ b/src/torrent/download.h
@@ -149,6 +149,10 @@ public:
   void                set_connection_type(ConnectionType t);
 
   enum HeuristicType {
+    HEURISTICS_UPLOAD_LEECH              = 0,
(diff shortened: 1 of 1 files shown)


rakshasa:
It should probably be removed and only use the enum from `torrent/download/choke_queue.h`.

To avoid unnecessary includes of that header, make it an enum in a new file `torrent/download/types.h`.

PR #761, opened 2026-06-09
[WIP] Enable curl connection reuse with SocketManager tracking

trim21:
## Summary

This PR enables curl connection reuse and adds proper SocketManager tracking of curl sockets throughout their lifecycle, eliminating repeated SSL handshake overhead for HTTP tracker announces.

### Connection reuse
- Set `CURLOPT_FORBID_REUSE=0` and `CURLOPT_SSL_SESSIONID_CACHE=1` so curl reuses established connections across tracker announces
- Removes the TODO comment noting connection reuse was disabled due to libcurl API concerns, which are now addressed

### Socket lifecycle tracking
- **CLOSESOCKETFUNCTION callback** — SocketManager is notified when curl purges cached connections, so quotas are properly managed
- **`mark_stream_event_inactive` for POLL_REMOVE** — idle/cached connections stay registered in SocketManager (marked inactive) rather than being fully unregistered, allowing reuse under quota limits without re-registration overhead
- **Connection cache limits** — `set_max_cache_connections(total_size/2)` in `ThreadNet`, leaving headroom for new active connections beyond the idle cache

### Error handling
- **`event_error` simplified** — delegates cleanup to curl via `CLOSESOCKETFUNCTION` instead of manual fd management, avoiding double-close races
- **Graceful `~CurlSocket()`** — replaces assertion crash with warning + close when socket is unexpectedly open at destruction
- **`close_socket` handles idle cached sockets** — graceful path for sockets removed from epoll but kept in socket_map

## Files changed
- `src/net/curl_get.cc` — connection reuse opts, CLOSESOCKETFUNCTION
- `src/net/curl_socket.cc` — POLL_REMOVE inactive tracking, event_error, close_socket, destructor
- `src/net/thread_net.cc` — cache connection limit
- `configure.ac` — libcurl >= 7.21.7 (for `CURLOPT_CLOSESOCKETFUNCTION`)


rakshasa:
See https://github.com/curl/curl/issues/21779 why this won't work.

trim21:
yeah, you are right. I encounter this.

PR #762, opened 2026-06-09
Fixed http tracker retry timeouts.

rakshasa:


trim21:
I think we should also have some kind of tracker announce concurrency limitations?

rakshasa:
The HTTP stack should be responsible for that, perhaps the max concurrent requests value is too high.

trim21:
> The HTTP stack should be responsible for that, perhaps the max concurrent requests value is too high.

network.http.max_total_connections limits simultaneous TCP connections (CURLMOPT_MAX_TOTAL_CONNECTIONS), not the number of queued requests. The CurlStack vector has no size limit, and do_timeout() dispatches all tracker events at once — with 5000 torrents every one of them gets pushed into curl's queue regardless of how many connections are open. A tracker-level limit like trackers.max_active paces the dispatch rate before requests ever reach curl, so the vector stays small and the event loop doesn't get overwhelmed by thousands of queued handles.

rakshasa:
I don't really like to add that kind of logic to the trackers as that adds a lot of complexity to already convoluted code, and callbacks should be fast enough to handle adding 5000+ to CurlStack without issue.

Seems the vector should be changed to a map, or unordered_map (though we should shrink it on a scheduled task when it shrinks if it is too large.

CurlStack used to limit max connections itself iirc, so the code makes it easy to add that back in.

For the to-be-activated list, use a deque and keep an iterator in CurlGet, and clear the value (ptr to CurlGet) when it is closed before activating.

PR #764, opened 2026-06-09
fix: scheduler O(N) make_heap to O(log N) via SchedulerHandle lazy deletion

trim21:
## Summary

Replace `make_heap()` (O(N)) in `erase()`/`update_wait_until()` with lazy deletion via `SchedulerHandle` — an intermediate heap-owned struct.

## Design

- **`SchedulerHandle`** — struct with `entry`, `scheduler`, `time`; owned by `unique_ptr` in `Scheduler::m_heap`
- **`Scheduler`** — owns `std::vector` (composition), one-liner `next_timeout(timeout)` handles empty/clamp internally
- **`erase()`** — clears `handle->entry` (O(1)), handle stays in heap for lazy cleanup
- **`update_wait_until()`** — cancels old handle + pushes new (O(log N))
- **`perform()`** — `std::ranges::pop_heap`, skips cancelled handles
- **`SchedulerEntry`** — removed `m_scheduler`, `scheduler()`, `time()`, `m_time`; now only `slot()`, `handle()`, `is_valid()`, `is_scheduled()`
- Removed: `sift_up/down`, `make_heap/push_heap` wrappers, `using base_type::`, vector inheritance

| Operation | Before | After |
|-----------|--------|-------|
| `erase()` | O(N) | O(1) |
| `update_wait_until()` | O(N) | O(log N) |
| `perform()` | O(log N) | O(log N) |

## Requires

- https://github.com/rakshasa/rtorrent/pull/1819

### DIFF
--- a/src/torrent/system/thread.cc
+++ b/src/torrent/system/thread.cc
@@ -278,9 +278,7 @@ Thread::event_loop() {
       instrumentation_update(instrumentation_enum(INSTRUMENTATION_POLLING_DO_POLL + m_instrumentation_index), 1);
 
       auto timeout = std::max(next_timeout(), std::chrono::microseconds(0));
-
-      if (!m_scheduler->empty())
-        timeout = std::min(timeout, m_scheduler->next_timeout());
+      timeout = m_scheduler->next_timeout(timeout);
 
       int event_count = m_poll->do_poll(timeout);
 
--- a/src/torrent/utils/scheduler.cc
+++ b/src/torrent/utils/scheduler.cc
@@ -10,35 +10,33 @@
 
 namespace torrent::utils {
 
+static constexpr auto compare = [](const std::unique_ptr<SchedulerHandle>& a,
+                                   const std::unique_ptr<SchedulerHandle>& b) {
+  return a->time > b->time;
+};
+
 SchedulerEntry::~SchedulerEntry() {
   assert(!is_scheduled() && "SchedulerEntry::~SchedulerEntry() called on a scheduled item.");
-  assert(m_time == time_type{} && "SchedulerEntry::~SchedulerEntry() called on an item with a time.");
 
   m_slot = nullptr;
-  m_scheduler = nullptr;
-  m_time = time_type{};
-}
-
-
-inline void
-Scheduler::make_heap() {
-  std::make_heap(begin(), end(), [](const SchedulerEntry* a, const SchedulerEntry* b) {
-      return a->time() > b->time();
-    });
-}
-
-inline void
-Scheduler::push_heap() {
-  std::push_heap(begin(), end(), [](const SchedulerEntry* a, const SchedulerEntry* b) {
-      return a->time() > b->time();
-    });
 }
 
 Scheduler::time_type
-Scheduler::next_timeout() const {
-  assert(!empty());
+Scheduler::next_timeout(Scheduler::time_type max_timeout) {
+  while (!m_heap.empty() && m_heap.front()->entry == nullptr) {
+    std::ranges::pop_heap(m_heap, compare);
+    m_heap.pop_back();
+  }
 
-  return std::max(front()->time() - m_cached_time, Scheduler::time_type());
+  if (m_heap.empty())
+    return max_timeout;
+
+  auto timeout = m_heap.front()->time - m_cached_time;
+
+  if (timeout >= max_timeout)
+    return max_timeout;
+
+  return std::max(timeout, Scheduler::time_type());
 }
 
 // We can't make erase/update part of SchedulerItem in case another thread tries to call the
@@ -55,21 +53,20 @@ Scheduler::erase(SchedulerEntry* entry) {
   if (!entry->is_valid())
     throw torrent::internal_error("Scheduler::erase(...) called on an invalid entry.");
 
-  if (entry->scheduler() != this)
+  if (entry->m_handle->scheduler != this)
     throw torrent::internal_error("Scheduler::erase(...) called on an entry that is in another scheduler.");
 
-  auto itr = std::find_if(begin(), end(), [entry](const SchedulerEntry* e) {
-      return e == entry;
-    });
+  entry->m_handle->entry = nullptr;
+  entry->set_handle(nullptr);
+}
 
-  if (itr == end())
-    throw torrent::internal_error("Scheduler::erase(...) could not find item in queue.");
+void
+Scheduler::push_entry(SchedulerEntry* entry, time_type time) {
+  auto handle = std::make_unique<SchedulerHandle>(SchedulerHandle{entry, this, time});
+  entry->set_handle(handle.get());
 
-  entry->set_scheduler(nullptr);
-  entry->set_time(Scheduler::time_type{});
-
-  base_type::erase(itr);
-  make_heap();
+  m_heap.push_back(std::move(handle));
+  std::ranges::push_heap(m_heap, compare);
 }
 
 void
@@ -88,11 +85,7 @@ Scheduler::wait_until(SchedulerEntry* entry, Scheduler::time_type time) {
   if (entry->is_scheduled())
     throw torrent::internal_error("Scheduler::wait_until(...) called on an already scheduled entry.");
 
-  entry->set_scheduler(this);
-  entry->set_time(time);
-
-  base_type::push_back(entry);
-  push_heap();
+  push_entry(entry, time);
 }
 
 void
@@ -125,19 +118,15 @@ Scheduler::update_wait_until(SchedulerEntry* entry, Scheduler::time_type time) {
     throw torrent::internal_error("Scheduler::update_wait(...) called on an invalid entry.");
 
   if (entry->is_scheduled()) {
-    if (entry->scheduler() != this)
+    if (entry->m_handle->scheduler != this)
       throw torrent::internal_error("Scheduler::update_wait(...) called on an entry that is in another scheduler.");
 
-    entry->set_time(time);
-    make_heap();
+    entry->m_handle->entry = nullptr;
+    push_entry(entry, time);
     return;
   }
 
-  entry->set_scheduler(this);
-  entry->set_time(time);
-
-  base_type::push_back(entry);
-  push_heap();
+  push_entry(entry, time);
 }
 
 void
@@ -158,17 +147,16 @@ Scheduler::update_wait_for_ceil_seconds(SchedulerEntry* entry, Scheduler::time_t
 
 void
 Scheduler::perform(Scheduler::time_type current_time) {
-  while (!empty() && base_type::front()->time() <= current_time) {
-    auto entry = base_type::front();
+  while (!m_heap.empty() && m_heap.front()->time <= current_time) {
+    std::ranges::pop_heap(m_heap, compare);
+    auto handle = std::move(m_heap.back());
+    m_heap.pop_back();
 
-    std::pop_heap(begin(), end(), [](const SchedulerEntry* a, const SchedulerEntry* b) {
-        return a->time() > b->time();
-      });
-    base_type::pop_back();
+    if (handle->entry == nullptr)
+      continue;
 
-    entry->set_scheduler(nullptr);
-    entry->set_time(Scheduler::time_type{});
-    entry->slot()();
+    handle->entry->set_handle(nullptr);
+    handle->entry->slot()();
   }
 }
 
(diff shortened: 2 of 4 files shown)


rakshasa:
This is too much container logic being put in Scheduler, it should be a separate class if we're adding this much logic.

However the erase and update functions could be optimized by using a pointer to a temporary object held by both SchedulerEntry and the heap, SchedulerEntry clears it when it wants to erase/update and it is left in the heap.

https://cppreference.com/cpp/algorithm/pop_heap

pop_head takes a range and comparitor, so we use that to pop all elements that have expired in perform.

rakshasa:
That change won't work, you need to have a temporary object between the heap and the scheduler entry.

rakshasa:
`Schedule` should be changed to inherit from `std::vector<std::unique_ptr<SchedulerHandle>>`.

```
struct SchedulerHandle {
  SchedulerEntry* entry{};
  Scheduler*      scheduler{}; // replaces SchedulerEntry::m_scheduler.
  time_type        time{};
};
```

When erasing/updating we clear the `entry` pointer and leave the `SchedulerItem` where it is.

Also remove `sift_up/down`, they are not needed and in perform use range-based pop_heap as that is as fast as can be.

rakshasa:
There's no need to make SchedulerHandle a complex class, it adds unnessesary complexity when the struct will never be exposed to users.

trim21:
> There's no need to make SchedulerHandle a complex class, it adds unnessesary complexity when the struct will never be exposed to users.

I try to avoid the need to update rtorrent, the struct/smart ptr need a time() method.

trim21:
can we consider that merge 2 repo again? I have 2 change but it cause 4 PR which cause too much trouble when contributing. if there are no another client that use libtorrent, this just add complex on development and distribution with minimal benefit

rakshasa:
`m_scheduler.front()->time()` this code should be replaced by a member function in Scheduler.

In fact all the `using base_type::` should be removed.

src/torrent/utils/scheduler.cc
```diff
@@ -10,35 +10,24 @@
...
-      return a->time() > b->time();
-    });
-}
-
 Scheduler::time_type
 Scheduler::next_timeout() const {
   assert(!empty());
 
-  return std::max(front()->time() - m_cached_time, Scheduler::time_type());
+  return std::max(front()->time - m_cached_time, Scheduler::time_type());
```
rakshasa: Now that we got potentially orphaned handles, remove them first.

This requires changing the interface to: `Scheduler::next_timeout(Scheduler::time_type max_timeout)`

src/torrent/utils/scheduler.h
```diff
@@ -61,25 +59,25 @@ class LIBTORRENT_EXPORT SchedulerEntry {
   ~SchedulerEntry();
 
   bool                is_valid() const     { return m_slot != nullptr; }
-  bool                is_scheduled() const { return m_scheduler != nullptr; }
+  bool                is_scheduled() const { return m_handle != nullptr; }
 
   slot_type&          slot()               { return m_slot; }
-  Scheduler*          scheduler() const    { return m_scheduler; }
+  Scheduler*          scheduler() const    { return m_handle ? m_handle->scheduler : nullptr; }
   time_type           time() const         { return m_time; }
```
rakshasa: Don't think we need `time()/scheduler()`.

trim21:
inherit from std::vector<std::unique_ptr<SchedulerHandle>> expose too many API from vector, I change it to own a vector as m_heap

src/torrent/utils/scheduler.cc
```diff
@@ -10,35 +10,40 @@
...
+    m_heap.pop_back();
+  }
 
-inline void
-Scheduler::make_heap() {
-  std::make_heap(begin(), end(), [](const SchedulerEntry* a, const SchedulerEntry* b) {
-      return a->time() > b->time();
-    });
-}
+  assert(!m_heap.empty() && "Scheduler::front_time() called on an empty scheduler — caller must check empty() first");
```
rakshasa: `front_time()` is a mess, and not used except a specific use-case in rtorrent.

Use `next_timeout()` there and remove this function.

src/torrent/utils/scheduler.cc
```diff
@@ -10,35 +10,40 @@
...
-Scheduler::next_timeout() const {
-  assert(!empty());
+Scheduler::next_timeout(Scheduler::time_type max_timeout) {
+  while (!m_heap.empty() && m_heap.front()->time <= max_timeout && m_heap.front()->entry == nullptr) {
+    std::ranges::pop_heap(m_heap, compare);
+    m_heap.pop_back();
+  }
+
+  if (m_heap.empty())
+    return Scheduler::time_type();
```
rakshasa: When empty it should return max_timeout, it's _not_ meant to be a replacement for m_cached_time.

src/torrent/utils/scheduler.h
```diff
@@ -2,29 +2,30 @@
 #define TORRENT_UTILS_SCHEDULER_H
 
 #include <atomic>
+#include <cassert>
```
rakshasa: Assert should never be in a header.

src/torrent/utils/scheduler.h
```diff
@@ -2,29 +2,30 @@
...
 #include <thread>
 #include <vector>
 #include <torrent/common.h>
 
 namespace torrent::utils {
 
-class LIBTORRENT_EXPORT Scheduler : public std::vector<SchedulerEntry*> {
+struct SchedulerHandle {
+  class SchedulerEntry* entry{};
+  class Scheduler*      scheduler{};
```
rakshasa: Why is `class` there?

src/torrent/utils/scheduler.h
```diff
@@ -45,41 +46,35 @@ class LIBTORRENT_EXPORT Scheduler : public std::vector<SchedulerEntry*> {
...
 
 private:
-  void                make_heap();
-  void                push_heap();
-
-  std::atomic<std::thread::id> m_thread_id;
-  time_type                    m_cached_time{};
+  std::atomic<std::thread::id>                m_thread_id;
+  time_type                                   m_cached_time{};
+  std::vector<std::unique_ptr<SchedulerHandle>> m_heap;
```
rakshasa: Add a using for the vector type.

src/tracker/tracker_controller.cc
```diff
@@ -63,17 +63,20 @@ TrackerController::is_scrape_queued() const {
 
 int64_t
 TrackerController::next_timeout() const {
-  return m_task_timeout.time().count();
+  return m_task_timeout.handle() ? m_task_timeout.handle()->time.count() : 0;
```
rakshasa: This should be a SchedulerEntry function.

rakshasa (commented):
Please put more thought into making clean interfaces, there's too much here that would have been much cleaner with just a bit of rearranging.

src/torrent/system/thread.cc
```diff
@@ -280,7 +280,7 @@ Thread::event_loop() {
       auto timeout = std::max(next_timeout(), std::chrono::microseconds(0));
 
       if (!m_scheduler->empty())
-        timeout = std::min(timeout, m_scheduler->next_timeout());
+        timeout = std::min(timeout, m_scheduler->next_timeout(m_cached_time.load()));
```
rakshasa: This misses the point, it should be:

```
auto timeout = std::max(next_timeout(), std::chrono::microseconds(0));

timeout = m_scheduler->next_timeout(timeout);
```

src/torrent/utils/scheduler.cc
```diff
@@ -10,35 +10,28 @@
...
+  while (!m_heap.empty() && m_heap.front()->time <= max_timeout && m_heap.front()->entry == nullptr) {
+    std::ranges::pop_heap(m_heap, compare);
+    m_heap.pop_back();
+  }
 
-  return std::max(front()->time() - m_cached_time, Scheduler::time_type());
+  if (m_heap.empty())
+    return max_timeout;
+
+  return std::max(m_heap.front()->time - m_cached_time, Scheduler::time_type());
```
rakshasa: This doesn't consider `max_timeout`, also use an if instead of std::max to make things clearer.

src/torrent/utils/scheduler.h
```diff
@@ -61,25 +63,23 @@ class LIBTORRENT_EXPORT SchedulerEntry {
...
 
   bool                is_valid() const     { return m_slot != nullptr; }
-  bool                is_scheduled() const { return m_scheduler != nullptr; }
+  bool                is_scheduled() const { return m_handle != nullptr; }
 
   slot_type&          slot()               { return m_slot; }
-  Scheduler*          scheduler() const    { return m_scheduler; }
-  time_type           time() const         { return m_time; }
+  SchedulerHandle*    handle() const       { return m_handle; }
+  time_type           time() const         { return m_handle ? m_handle->time : time_type{}; }
```
rakshasa: We should get rid of `time()` also.

src/torrent/utils/scheduler.cc
```diff
@@ -10,35 +10,37 @@
...
-  m_scheduler = nullptr;
-  m_time = time_type{};
 }
 
+Scheduler::time_type
+Scheduler::next_timeout(Scheduler::time_type max_timeout) {
+  while (!m_heap.empty() && m_heap.front()->time <= max_timeout && m_heap.front()->entry == nullptr) {
+    std::ranges::pop_heap(m_heap, compare);
+    m_heap.pop_back();
+  }
```
rakshasa, line 29: Don't check max_timeout here, purge all the orphaned entries.

src/torrent/utils/scheduler.cc
```diff
@@ -10,35 +10,37 @@
...
 
-Scheduler::time_type
-Scheduler::next_timeout() const {
-  assert(!empty());
+  if (rel_time < max_timeout){
+    return rel_time;
+  }
 
-  return std::max(front()->time() - m_cached_time, Scheduler::time_type());
+  return max_timeout;
```
rakshasa: This produces the wrong result.

```
auto timeout = m_heap.front()->time - m_cached_time;

if (timeout >= max_timeout)
  return max_timeout;

return std::max(timeout, Scheduler::time_type());
```

Use max/min when you're restricting a type value, and if branch when it is logic related.

src/torrent/utils/scheduler.cc
```diff
@@ -125,19 +117,24 @@ Scheduler::update_wait_until(SchedulerEntry* entry, Scheduler::time_type time) {
...
 
-  entry->set_scheduler(this);
-  entry->set_time(time);
+  auto handle = std::make_unique<SchedulerHandle>(SchedulerHandle{entry, this, time});
+  entry->set_handle(handle.get());
 
-  base_type::push_back(entry);
-  push_heap();
+  m_heap.push_back(std::move(handle));
+  std::ranges::push_heap(m_heap, compare);
```
rakshasa: This same piece of code is in three places, make it a member function.

src/torrent/utils/scheduler.cc
```diff
@@ -158,16 +155,16 @@ Scheduler::update_wait_for_ceil_seconds(SchedulerEntry* entry, Scheduler::time_t
...
-      });
-    base_type::pop_back();
+    if (handle->entry == nullptr)
+      continue;
 
-    entry->set_scheduler(nullptr);
-    entry->set_time(Scheduler::time_type{});
+    auto entry = handle->entry;
+    entry->set_handle(nullptr);
     entry->slot()();
```
rakshasa: entry is not needed here, just use handle->entry the whole way.

src/tracker/tracker_controller.cc
rakshasa: This still needs to be a SchedulerEntry method:

`return m_task_timeout.time_or_zero().count()`

And why are you exposing handle, it's internal to scheduler, the outside should never see it.

src/tracker/tracker_controller.cc
```diff
@@ -87,7 +90,10 @@ TrackerController::seconds_to_next_timeout() const {
 
 uint32_t
 TrackerController::seconds_to_next_scrape() const {
-  auto timeout = m_task_scrape.time() - this_thread::cached_time();
+  if (!m_task_scrape.is_scheduled())
+    return 0;
+
+  auto timeout = m_task_scrape.handle()->time - this_thread::cached_time();
```
rakshasa: These if statements are unnecessary, they are used for display only and 0 is returned when not scheduled so use time_or_zero with cachetime and std::max.

PR #770, opened 2026-06-12
fix: set epoll timeout=0 only when has_interrupt_callbacks

trim21:


rakshasa:
This is incorrect, callbacks always interrupt a polling thread.

Normal callbacks get executed once or twice (depending on thread type) per poll loop, while interrupt callbacks also get interlaced between read/write events.

PR #778, opened 2026-06-15
Add support for RISC-V

andreas-schwab:


rakshasa:
It should be a separate case, and match both `riscv32*` and `riscv64*` explicitly and make them 64 bytes just to be safe.

PR #784, opened 2026-06-18
perf: skip sync_chunks early when queue is empty

trim21:
Avoid unnecessary statvfs calls in the periodic sync path by returning
early from ChunkList::sync_chunks when m_queue is empty.

The free_diskspace check (which calls statvfs) is only needed when there
are dirty chunks to flush, but the previous code evaluated it unconditionally.
This is especially wasteful for seeding torrents where m_queue is always
empty since no data is written.

## Background

`sync_chunks()` is called every 30s via `ChunkManager::periodic_sync()`
for *all* downloads in the ChunkManager, including seeding/stopped torrents.
Before this change, `statvfs` was invoked for every download on every tick,
even when there were no dirty chunks to flush.

## Change

Add an early return at the start of `ChunkList::sync_chunks()` when
`m_queue` is empty — nothing to sync means nothing to check.

### DIFF
--- a/src/data/chunk_list.cc
+++ b/src/data/chunk_list.cc
@@ -250,6 +250,13 @@ uint32_t
 ChunkList::sync_chunks(sync_flags flags) {
   LT_LOG_THIS(DEBUG, "Sync chunks: flags:%#x.", flags);
 
+  // An empty queue means no chunks were written since the last sync,
+  // so there is nothing to flush. Skipping early avoids unnecessary
+  // statvfs calls in the free_diskspace check below - especially
+  // optimize for seeding torrents where no data is being written.
+  if (m_queue.empty())
+    return 0;
+
   Queue::iterator split;
 
   if ((flags & sync_all))


rakshasa:
This causes `tests/ci-low-diskspace` to fail every time.

Instead of the above solution, `FileList::free_diskspace()` should take a vector as argument containing a cache of values.

This should reduce the number of `statvfs` calls to just a one per mount point per sync_all / try_free_memory.

trim21:
where is this `tests/ci-low-diskspace` ?

rakshasa:
`rtorrent-docker` repo

trim21:
🤔 shouldn't we fix tests/ci-low-diskspace and don't use a seeding torrent to test the behavoir?

rakshasa:
`feature/cache-free-diskspace`

This optimizes free diskspace checks, and the low-diskspace test does fail sometimes and needs fixing.

rakshasa:
Could you update this to the latest master?

trim21:
I think it's fine we we just call statvfs once for each mountpoint in each tick

rakshasa:
I added it to my curl fix branch, so it'll get added eventually.

PR #786, opened 2026-06-18
bitfield: empty Bitfield must have begin() == end()

xirvik:
A Bitfield with its size set but data unallocated (m_data == nullptr) -- size set via set_size_bits() before allocate(), or after unallocate() while the size is retained -- returned end() = m_data.get() + size_bytes(), i.e. the range [nullptr, nullptr + N). Iterating it dereferences null. Make end() == begin() (nullptr) when m_data is null so an unallocated bitfield is an empty range, per std-container semantics. Fixes a SIGSEGV in transform_to_hex_str() reachable from rtorrent d.bitfield via d.multicall2 (ruTorrent/pymedusa). 

Suggested by @rakshasa in rtorrent#1836.

### DIFF
--- a/src/torrent/bitfield.h
+++ b/src/torrent/bitfield.h
@@ -63,13 +63,8 @@ public:
 
   iterator            begin()                       { return m_data.get(); }
   const_iterator      begin() const                 { return m_data.get(); }
-  // std-container invariant: an empty bitfield (no allocated data) must have
-  // begin() == end(). When m_data is null, size_bytes() can still be > 0 (size set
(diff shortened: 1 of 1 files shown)


rakshasa:
For future PRs, don't add unnecessary (or long) comments. Only rarely does a piece of code need any comments, and if they do it's usually better to name the variables/functions better.

xirvik:
> For future PRs, don't add unnecessary (or long) comments. Only rarely does a piece of code need any comments, and if they do it's usually better to name the variables/functions better.

Noted

PR #797, opened 2026-06-26
O_CLOEXEC everywhere

fffe:
Try to prevent leaking fds to child processes by atomically setting O_CLOEXEC.

Related to https://github.com/rakshasa/rtorrent/issues/1848

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -67,7 +67,7 @@ if test "x$use_kqueue" != "xyes" && test "x$use_epoll" != "xyes"; then
 fi
 
 AX_PTHREAD
-AC_CHECK_FUNCS([fallocate posix_fallocate])
+AC_CHECK_FUNCS([fallocate posix_fallocate accept4 pipe2 epoll_create1 kqueue1 inotify_init1])
 
 TORRENT_ENABLE_CUSTOM_STACK_SIZE
 
--- a/src/data/socket_file.cc
+++ b/src/data/socket_file.cc
@@ -8,6 +8,7 @@
 #include <unistd.h>
 
 #include "torrent/exceptions.h"
+#include "torrent/net/fd.h"
 #include "torrent/utils/file_stat.h"
 #include "torrent/utils/log.h"
 
@@ -30,7 +31,7 @@ SocketFile::open(const std::string& path, int prot, int flags, mode_t mode) {
   else
     throw internal_error("torrent::SocketFile::open(...) Tried to open file with no protection flags");
 
-  fd_type fd = ::open(path.c_str(), flags, mode);
+  fd_type fd = fd_open_file(path, flags, mode);
 
   if (fd == invalid_fd)
     return false;
--- a/src/torrent/net/fd.cc
+++ b/src/torrent/net/fd.cc
@@ -7,6 +7,19 @@
 #include <netinet/tcp.h>
 #include <unistd.h>
 
+#if defined(USE_EPOLL)
+#include <sys/epoll.h>
+#elif defined(USE_KQUEUE)
+#include <sys/time.h>
+#include <sys/event.h>
+#else
+#error "Must enable at least one of either kqueue or epoll"
+#endif
+
+#ifdef USE_INOTIFY
+#include <sys/inotify.h>
+#endif
+
 #include "torrent/exceptions.h"
 #include "torrent/net/socket_address.h"
 #include "torrent/utils/log.h"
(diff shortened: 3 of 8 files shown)


rakshasa:
This isn't needed with posix_spawn and addclosefrom_np.

fffe:
addclosefrom_np doesn't exist in musl libc (or on openbsd/netbsd)

Dunno if openbsd and netbsd matter, but there are a lot of alpine container images.

rakshasa:
At this point might as well consider finishing https://github.com/rakshasa/shm-ipc and using that to split ExecFile into its own process.

src/torrent/utils/directory_events.cc
```diff
@@ -57,7 +57,11 @@ directory_events::open() {
   errno = 0;
 
 #ifdef USE_INOTIFY
+#ifdef HAVE_INOTIFY_INIT1
+  m_fileDesc = inotify_init1(IN_CLOEXEC);
+#else
   m_fileDesc = inotify_init();
+#endif
```
rakshasa: Replace with `fd_inotify_init()` so all the CLOEXEC stuff is in fd.cc

src/data/socket_file.cc
```diff
@@ -30,7 +30,7 @@ SocketFile::open(const std::string& path, int prot, int flags, mode_t mode) {
   else
     throw internal_error("torrent::SocketFile::open(...) Tried to open file with no protection flags");
 
-  fd_type fd = ::open(path.c_str(), flags, mode);
+  fd_type fd = ::open(path.c_str(), flags | O_CLOEXEC, mode);
```
rakshasa: This should be changed to use `fd_open`.

src/torrent/system/poll_epoll.cc
```diff
@@ -154,7 +154,11 @@ Poll::create() {
   if (socket_open_max == -1)
     throw internal_error("Poll::create() : sysconf(_SC_OPEN_MAX) failed : " + std::string(std::strerror(errno)));
 
+#ifdef HAVE_EPOLL_CREATE1
+  int fd = epoll_create1(EPOLL_CLOEXEC);
+#else
   int fd = epoll_create(socket_open_max);
+#endif
```
rakshasa: Add an `fd_epoll_create`

rakshasa:
I'd rather get 1.0 out before adding shm-ipc, so might as well fix this.

src/torrent/system/poll_epoll.cc
rakshasa: Also need to use ``kqueue1`, perhaps we should have an `fd_poll_create()` function.

src/data/socket_file.cc
fffe: that's a bigger change. fd_open is (confusingly) for sockets, and despite the SocketFile name this is a regular file.

what do you want to do there?

rakshasa: I guess `fd_open_file()` would make sense then.

src/torrent/net/fd.h
```diff
@@ -23,8 +24,18 @@ constexpr bool  fd_valid_flags(fd_flags flags);
...
 void            fd_open_socket_pair(int& fd1, int& fd2) LIBTORRENT_EXPORT;
+#ifdef USE_EPOLL
+int             fd_epoll_create(int size) LIBTORRENT_EXPORT;
+#endif
+#ifdef USE_INOTIFY
+int             fd_inotify_init() LIBTORRENT_EXPORT;
+#endif
+#ifdef USE_KQUEUE
+int             fd_kqueue() LIBTORRENT_EXPORT;
+#endif
```
rakshasa: No need to pollute the headers, just throw internal_error if called.

Also, I'd prefer `fd_open_epoll/kqueue/inotify` to be consistent.

configure.ac
```diff
@@ -67,7 +67,19 @@ if test "x$use_kqueue" != "xyes" && test "x$use_epoll" != "xyes"; then
...
+  AC_CHECK_FUNCS([epoll_create1])
+fi
+
+if test "x$use_kqueue" = "xyes"; then
+  AC_CHECK_FUNCS([kqueue1])
+fi
+
+if test "x$use_inotify" = "xyes"; then
+  AC_CHECK_FUNCS([inotify_init1])
+fi
```
rakshasa: This shouldn't pollute configure.ac, you can either just check for them all above, or add them to the scripts.

src/torrent/net/fd.cc
```diff
@@ -7,6 +7,19 @@
...
+#endif
+
+#ifdef USE_INOTIFY
+#include <sys/inotify.h>
+#endif
+
+#ifdef USE_KQUEUE
+#include <sys/time.h>
+#include <sys/event.h>
+#endif
```
rakshasa: epoll/kqueue checks should be if/elif/else error.

configure.ac
fffe: Ah okay. I figured no point in testing for inotify_init1 if inotify_init isn't present, but I'll put it back how it was.

src/torrent/net/fd.cc
fffe: do you mean if one of the fd_* stubs is called? otherwise I don't understand what the error case is for this bit.

rakshasa: No, I mean the header includes and the ifdef's.

While there are sanity checks for epoll/kqueue elsewhere, it still looks bad just putting them in separate ifdef blocks.

fffe:
I think I got what you wanted, and added SOCK_CLOEXEC on one I missed in net/curl_socket.

rakshasa:
Looks good, probably should make a quick .17 release.

rakshasa:
```
#30 16.03 
#30 16.03 torrent/utils:
#30 16.03 
#30 16.03 torrent/utils::test_extents:
#30 16.03 torrent/utils::test_extents::test_basic
#30 16.03 
#30 16.03 Caught Segmentation fault, dumping stack:
#30 16.03 
#30 16.03 Stack dump not enabled.
#30 16.03 
```

rakshasa:
Those issues usually aren't related to the tests themselves, but the custom libtorrent test framework.

rakshasa:
`fffe-cloexec_fixup`

PR #798, opened 2026-06-27
Add libcurl limits on maximum announce size and redirect protocols

anthonyryan1:
Healthy tracker announces are always going to be quite small, and the default lack of upper bound on content length means that curl can allocate as much memory as the tracker sends over the network.

This can result in OOM killer going after rtorrent when trackers deploy buggy releases, and return an infinte loop of the same peers or if the announce URL is pointed at a large static file rather than a tracker.

We can mitigate that by telling curl to abort any requests with responses over 1 MiB, which is a very high limit relative to the average announce response, and better than unbounded.

The other new limit, is restricting which protocols curl will follow as part of CURLOPT_FOLLOWLOCATION. Old versions of curl would follow anything including file:// url's, while more recent versions include ftp by default.

I expect no tracker has ever actually redirected rtorrent to connect to an FTP server, but it would probably be preferable it wasn't possible.


rakshasa:
These should probably be added as options to HttpGet, and tracker requests should have different limits than rtorrent downloading torrent files.

rakshasa:
There's now a 1mb limit on tracker requests, and 15mb limit on torrents.

PR #801, opened 2026-07-03
Added max http request size and limited redirect protocols to http/https.

rakshasa:


rakshasa:
Closes https://github.com/rakshasa/libtorrent/pull/798

PR #805, opened 2026-07-04
Fix encryption_require bitmask falsely matching allow_incoming

j-c-m:
found this when sendfile() testing

encryption_require is 0x3 (allow_incoming | try_outgoing). Bitwise checks against it also matched allow_incoming alone, causing outgoing connections to use encrypted handshakes when only allow_incoming was configured. Add NetworkConfig helper functions and use exact-match semantics for require-specific behavior.

### DIFF
--- a/src/protocol/encryption_info.h
+++ b/src/protocol/encryption_info.h
@@ -37,12 +37,28 @@
 #ifndef LIBTORRENT_PROTOCOL_ENCRYPTION_H
 #define LIBTORRENT_PROTOCOL_ENCRYPTION_H
 
+#include <cstdint>
+
 #include "utils/rc4.h"
 
 namespace torrent {
 
 class EncryptionInfo {
 public:
+  static constexpr uint32_t option_none             = 0;
+  static constexpr uint32_t option_allow_incoming   = 0x1;
+  static constexpr uint32_t option_try_outgoing     = 0x2;
+  static constexpr uint32_t option_require          = 0x3;
+  static constexpr uint32_t option_require_RC4      = 0x4;
+  static constexpr uint32_t option_enable_retry     = 0x8;
+  static constexpr uint32_t option_prefer_plaintext = 0x10;
+  // Internal to libtorrent.
+  static constexpr uint32_t option_retrying         = 0x40;
+
+  static bool         try_outgoing(uint32_t opts);
+  static bool         is_required(uint32_t opts);
+  static bool         allow_incoming(uint32_t opts);
+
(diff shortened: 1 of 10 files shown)


src/torrent/runtime/network_config.h
```diff
@@ -31,6 +31,18 @@ class LIBTORRENT_EXPORT NetworkConfig {
...
+    return (opts & encryption_try_outgoing) != 0;
+  }
+
+  static bool         encryption_is_required(uint32_t opts) {
+    return (opts & encryption_require) == encryption_require;
+  }
+
+  static bool         encryption_accepts_incoming_encrypted(uint32_t opts) {
+    return (opts & encryption_allow_incoming) != 0;
+  }
```
rakshasa: I've been wanting to remove the encryption flags from NetworkConfig, not add more.

RTorrent should be changed to use `torrent::option_find_string` only, and the above static flags moved to e.g. `src/protocol/encryption_info.h`

src/protocol/encryption_info.h
```diff
@@ -37,12 +37,39 @@
...
+    return (opts & option_try_outgoing) != 0;
+  }
+
+  static bool         is_required(uint32_t opts) {
+    return (opts & option_require) == option_require;
+  }
+
+  static bool         accepts_incoming_encrypted(uint32_t opts) {
+    return (opts & option_allow_incoming) != 0;
+  }
```
rakshasa: Define these as inline functions below the class, as we want to keep the class declaration clean.

j-c-m: I wasn't really sure on the style, because they could also be a single line, but long. I moved them to inline below the class for now.

src/protocol/encryption_info.h
```diff
@@ -37,12 +37,28 @@
...
+  static constexpr uint32_t option_require          = 0x3;
+  static constexpr uint32_t option_require_RC4      = 0x4;
+  static constexpr uint32_t option_enable_retry     = 0x8;
+  static constexpr uint32_t option_prefer_plaintext = 0x10;
+  // Internal to libtorrent.
+  static constexpr uint32_t option_retrying         = 0x40;
+
+  static bool         uses_outgoing(uint32_t opts);
+  static bool         is_required(uint32_t opts);
+  static bool         accepts_incoming_encrypted(uint32_t opts);
```
rakshasa: `handshake_outgoing_uses_encryption`, `handshake_requires_encryption`, `handshake_incoming_accepts_encrypted` seem better.

rakshasa:
It's been too long since I looked at this code.

This change should obviously be in `src/protocol/handshake_encryption.h`.

PR #808, opened 2026-07-05
Encryption Refactor - was just a bitwise logic fix.

j-c-m:
Introduce EncryptionPolicy

Replace NetworkConfig encryption_options int bitmask with a four-field
EncryptionPolicy (Incoming, Outgoing, Cipher, Retry). Deprecates the
existing bitmask-based encryption_options API. Fix the
require/allow_incoming interaction bug.

companion: https://github.com/rakshasa/rtorrent/pull/1866


rakshasa:
Adding a new header seems over-kill, and I don't really like the current implementation in general.

There should be a cleaner, more user-friendly way of exposing these options to the user.

j-c-m:
Wouldn't that live more in rtorrent? I can see doing something like
```
protocol.encryption.mode.set = (off | default | on | require)
```

and then boolean options to get more advanced (presets would internally map to these) for the rest ie:
```
protocol.encryption.allow_incoming.set = 1
protocol.encryption.try_outgoing.set = 1
```

But wouldn't that mostly live in the rtorrent client and just map to flags here in libtorrent? Or am I thinking about the layering wrong?

rakshasa:
Separating incoming / outgoing makes users think about who connects to who, when really it is irrelevant.

User really selects from three preferences, and options you can set;

 * plaintext: allow encrypted handshakes, retry with encrypted handshake
   * disable-encrypted-handshakes option
 * default: allow plaintext, allow encrypted, prefer encrypted handshakes
   * prefer-encrypted-connections option
   * prefer-plaintext-handshakes (should this be inverted)
 * encrypted: reject encrypted handshakes, only allow encrypted

The `prefer-encrypted-connections` option is really not needed, since if you can have both incoming and outgoing plaintext connections there's really no reason to force a preference for encrypted. Your connections already show you as a BT user.

j-c-m:
right now this is the default - 

"protocol.encryption.set=allow_incoming,prefer_plaintext,enable_retry"

but with the bug in the current bit logic, allow_incoming actually forces try_outgoing and ignores prefer_plaintext.

j-c-m:
I did some bad git here, bear with me, going to untangle it

j-c-m:
So I just took this PR all the way, because I was in there. This is a breaking change that exposes a user-friendly configuration (protocol.encryption.mode.set), and much cleaner named granular options if a user is so inclined. These are the current presets (for the default - permissive - I copied the current default in main.cc)

```
protocol.encryption.mode.set = permissive   # default; most users
# Accept encrypted incoming peers, but connect outgoing in plain text.
# Prefer plaintext when a peer offers a choice after the encrypted handshake.
# If an outgoing handshake fails, retry once with the opposite encryption choice.

protocol.encryption.mode.set = enabled      # try encryption both ways
# Accept encrypted incoming peers and attempt encrypted outgoing connections.
# Negotiate cipher with the peer instead of preferring plaintext.
# Retry once on handshake failure by flipping the encryption attempt.

protocol.encryption.mode.set = strict       # encrypted-only
# Require MSE in both directions; reject plain handshakes.
# Require RC4 after negotiation; no plaintext fallback.
# No retry: failed encrypted handshakes are not retried as plain.

protocol.encryption.mode.set = disabled     # no encryption
# Plain incoming and outgoing only; no MSE handshake negotiation.
# No retry behavior for encryption-related handshake failures.
```

rakshasa:
This feels like it's going the wrong direction.

To make it simple and easy to use, and implement, we need to apply the same logic bidirectionally. And that makes sense from the user's network perspective.

And the implementation should reflect that, as in the same set of options should be applied for both incoming and outgoing.

```
protocol.encryption.set = deny
protocol.encryption.set = allow
protocol.encryption.set = prefer
protocol.encryption.set = require

protocol.encryption.handshake.set = deny
protocol.encryption.handshake.set = allow
protocol.encryption.handshake.set = require
```

First is full PE only.

Setting the first changes handshake to the default for that value, and restricts it.

Not allowed settings; allow/prefer+deny, require+deny/allow

Allowed settings; deny+require, ...

j-c-m:
Its as close as its ever been, I think it'll get there.

rakshasa:
The issue is your code adds different keywords/enums for incoming and outgoing.

With my suggestion it all collapses into one unified behavior.

j-c-m:
I am actively collapsing it. 

todo-
remove presets
merge incoming/outgoing
remove/hide retry (it would be implied in outgoing = allow) 

```
# --- Protocol encryption (PE handshake / optional RC4 stream) ---

# Preset mode (sets all granular options at once). Default at startup: permissive
#
#   disabled   - incoming=deny  outgoing=deny  stream=prefer  retry=0
#   permissive - incoming=allow outgoing=deny  stream=allow   retry=1
#   enabled    - incoming=allow outgoing=try   stream=prefer  retry=1
#   strict     - incoming=require outgoing=require stream=require retry=0
#
# protocol.encryption.mode.set = permissive

# Incoming (PE handshake acceptance):
#   deny    - accept only standard unencrypted BT handshakes
#   allow   - accept plain or PE handshakes
#   require - reject plain; require PE handshake
#
# protocol.encryption.incoming.set = allow

# Outgoing (PE handshake initiation):
#   deny    - send standard unencrypted BT handshake first
#   try     - send PE handshake first
#   require - require PE handshake (no plain fallback)
#
# protocol.encryption.outgoing.set = deny
# Stream (encryption after PE handshake; incoming peer chooses):
#
#   deny    - outgoing: offer handshake-only only
#             incoming:  require handshake-only in peer offer
#             outgoing accept: reject if peer chooses RC4
#
#   allow   - outgoing: offer handshake-only + RC4
#             incoming:  prefer handshake-only when peer offers both
#             outgoing accept: accept either
#
#   prefer  - outgoing: offer handshake-only + RC4
#             incoming:  prefer RC4 when peer offers both
#             outgoing accept: accept either
#
#   require - outgoing: offer RC4 only
#             incoming:  require RC4 in peer offer
#             outgoing accept: reject if peer chooses handshake-only
#
# Handshake-only = PE handshake, plaintext stream (peer list: H/h, detail: "handshake")
# RC4            = PE handshake + encrypted stream   (peer list: R/L, detail: "yes")
#
# protocol.encryption.stream.set = allow

# Retry on handshake failure:
#   0 - do not retry
#   1 - retry once, flipping plain <-> PE handshake path
#
# protocol.encryption.retry.set = 1
```

j-c-m:
testing...

```
# Default at startup: handshake=allow stream=allow
#
# Two independent axes. No preset modes, no separate retry setting.
# Retry is baked into handshake (outgoing-only, at most once).
# Stream applies only on the PE path, not on a plain-BT retry.
#
# Inspect effective policy:
#   protocol.encryption
#     -> handshake=allow stream=allow
#
# Handshake (PE handshake; symmetric for incoming and outgoing):
#   deny    - plain BT handshake only both ways; no retry
#   allow   - accept plain or PE inbound; PE first outbound with one plain retry
#   require - PE handshake required both ways; no retry
#
protocol.encryption.handshake.set = allow
#
# Stream (cipher negotiation after PE handshake succeeds):
#   deny    - offer handshake-only; require handshake-only (no RC4)
#   allow   - offer both; prefer handshake-only when choosing (incoming)
#   prefer  - offer both; prefer RC4 when choosing (incoming)
#   require - offer RC4 only; require RC4
#
# On outgoing PE connections you advertise per stream setting, but accept
# the peer's pick (unless deny/require). No stream-level retry.
#
protocol.encryption.stream.set = allow
```

j-c-m:
I purposely did the layers so rtorrent does the mapping to underlying more complex libtorrent layer, I can smash this all down if you want.

https://github.com/j-c-m/rtorrent/blob/5d78f10941d68e1e2d0722a19c3e84e2d061abb8/src/encryption_config.cc

j-c-m:
Also still need to inline/simplify/reduce/eliminate the helpers.

rakshasa:
You're still separating the behaviors in the code, there should be only these;

```
# Global State
stream_encryption = [deny, allow, prefer, require]
handshake_encryption = [deny, allow, prefer, require]

# Incoming State
stream_encryption
handshake_encryption

# Incoming Plaintext Connection
check if allows_plaintext_handshake
done

# Incoming Encrypted Connection
check if allows_encrypted_handshake
send negotation (offer):
  if require_encrypted_stream
  else if require_plaintext_stream // (currently missing in the handshake code)
  else if allows_encrypted_stream
  else internal_error
receive neg-reply:
  select based on stream_encryption or disconnect
done

# Outgoing State
stream_encryption
handshake_encryption
is_retrying = [false,true]

# Outcoming Connection (initial)
if prefer_encrypted_handshake
  initiate encrypted handshake
else
  initiate plaintext handshake

# Outcoming Connection (retry)
...

# Outcoming Plaintext Connection
check if allows_plaintext_handshake
done

# Outgoing Encrypted Connection
send negotation (offer):
  if require_encrypted_stream
  else if require_plaintext_stream // (currently missing in the handshake code)
  else
    offer plain+crypt
receive neg-reply:
  if selected_plaintext
    check if allows_plaintext_stream
  else if selected_crypto
    check if allows_encrypted_stream
done
```

j-c-m:
All that logic is covered in both directions in libtorrent in this PR, the configuration layer that lives in rtorrent simplifies to only those cases by unifying outgoing/incoming handshake.


Like I previously commented, I can smash this all down to libtorrent, but then libtorrent loses the flexibility to define handshake directionally.

I'll do that if that's how you want the layering to work, I just thought libtorrent flexibility had some benefit.

It still gets the functionality I want which is; handshake=required stream=none + sendfile .

rakshasa:
Required handshake, deny stream is possible with my suggestion. And it makes it apply uniformly in both directions.

Just try to implement it, it will become clear why I'm preferring this solution.

src/torrent/runtime/encryption_policy.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace torrent {
+
+class LIBTORRENT_EXPORT EncryptionPolicy {
+public:
+  enum class Handshake : uint8_t {
+    deny,
+    allow,
+    require,
+  };
```
rakshasa, line 25: Handshake needs a `prefer` to make the choice when doing outgoing connections. (`allow` means we first try with plaintext)

They should be a single enum class.

src/torrent/runtime/encryption_policy.h
```diff
@@ -0,0 +1,36 @@
...
+    deny,
+    allow,
+    prefer,
+    require,
+  };
+
+  Handshake handshake{Handshake::deny};
+  Stream    stream{Stream::prefer};
+
+  std::string to_string() const;
```
rakshasa, line 39: We don't have `to_string()` within libtorrent, converting between enums/flags and strings belong in `torrent/utils/option_strings`.

src/torrent/runtime/encryption_policy.h
```diff
@@ -0,0 +1,36 @@
...
+
+  enum class Stream : uint8_t {
+    deny,
+    allow,
+    prefer,
+    require,
+  };
+
+  Handshake handshake{Handshake::deny};
+  Stream    stream{Stream::prefer};
```
rakshasa: There's only one enum class.

Also those defaults are wrong, the default should be `allow` and `allow`.

src/protocol/handshake_manager.h
```diff
@@ -52,7 +53,7 @@ class HandshakeManager : private utils::unordered_vector<std::unique_ptr<Handsha
   HandshakeManager(const HandshakeManager&) = delete;
   HandshakeManager& operator=(const HandshakeManager&) = delete;
 
-  void                create_outgoing(const sockaddr* sa, DownloadMain* info, int encryptionOptions);
+  void                create_outgoing(const sockaddr* sa, DownloadMain* info, EncryptionPolicy policy, bool is_handshake_retry = false, bool is_stream_fallback = false);
```
rakshasa: Put those extra arguments in `policy`.

src/protocol/handshake_manager.cc
```diff
@@ -279,18 +283,41 @@ HandshakeManager::receive_failed(Handshake* ptr, int message, int error) {
...
+    || state == Handshake::READ_ENC_SKEY
+    || state == Handshake::READ_ENC_NEGOT
+    || state == Handshake::READ_ENC_PAD
+    || state == Handshake::READ_ENC_IA;
+
+  const bool stream_retry_eligible =
+    in_pe_negotiation
+    || error == Handshake::e_handshake_invalid_encryption
+    || error == Handshake::e_handshake_encryption_sync_failed
+    || error == Handshake::e_handshake_invalid_value;
```
rakshasa: All this is over-complicated.

We don't know why the connection got rejected, and we aren't smart enough to figure it out. So just retry if the policy says we should.

j-c-m:
It's simplified down, I added stream retries on outgoing so it performs the same as incoming, which was a little more complicated, testing now. 

```
# --- Protocol encryption (PE handshake / optional RC4 stream) ---
# Default at startup: handshake=allow stream=allow
#
# Handshake (PE handshake; symmetric for incoming and outgoing):
#   deny    - plain BT handshake only both ways; no retry
#   allow   - accept plain or PE inbound; PE first outbound with one plain retry
#   require - PE handshake required both ways; no retry
#
# protocol.encryption.handshake.set = allow
#
# Stream (cipher negotiation after PE handshake succeeds):
#   deny    - offer handshake-only; require handshake-only (no RC4)
#   allow   - incoming: prefer handshake-only if offered, else RC4
#             outgoing: offer handshake-only first, one PE retry with RC4-only
#   prefer  - outgoing: offer RC4 first, one PE retry offering both if that
#             fails; incoming: pick RC4 if offered, else handshake-only
#   require - offer RC4 only; require RC4
#
# Outgoing retry summary with handshake=allow:
#   stream=allow   -> PE (plain) -> PE (RC4) -> plain BT
#   stream=prefer  -> PE (RC4-only) -> PE (both) -> plain BT
#   stream=deny    -> PE (handshake-only) -> plain BT
#   stream=require -> PE (RC4-only) -> plain BT
#
# protocol.encryption.stream.set = allow
```

src/protocol/handshake.cc
```diff
@@ -994,20 +1027,21 @@ Handshake::event_write() {
...
-        // if connection fails, peer probably closed it because it was encrypted, so retry encrypted if enabled
-        if (!(m_encryption.options() & runtime::NetworkConfig::encryption_require))
-          m_encryption.set_retry(HandshakeEncryption::RETRY_PLAIN);
+        if (!m_is_outgoing_retry && m_encryption.policy().handshake == EncryptionPolicy::Handshake::allow)
+          m_encryption.add_retry(HandshakeEncryption::RETRY_PLAIN);
+
+        if (!m_is_stream_fallback
+            && (m_encryption.policy().stream == EncryptionPolicy::Stream::prefer
+                || m_encryption.policy().stream == EncryptionPolicy::Stream::allow))
+          m_encryption.add_retry(HandshakeEncryption::RETRY_STREAM_ALLOW);
```
rakshasa: Too complicated.

Why are you even including a stream retry?

The retry is only about handshake, the negotiation establishes if we use encrypted or plaintext stream. No retry there.

src/protocol/handshake.cc
```diff
@@ -430,32 +421,72 @@ Handshake::read_encryption_negotiation() {
...
     if (m_encryption.crypto() != HandshakeEncryption::crypto_rc4 && m_encryption.crypto() != HandshakeEncryption::crypto_plain)
       throw handshake_error(handshake_failed, e_handshake_invalid_encryption);
 
-    if ((m_encryption.options() & runtime::NetworkConfig::encryption_require_RC4) && (m_encryption.crypto() != HandshakeEncryption::crypto_rc4))
+    if (m_encryption.policy().stream == EncryptionPolicy::Stream::require
+        && (m_encryption.crypto() != HandshakeEncryption::crypto_rc4))
+      throw handshake_error(handshake_failed, e_handshake_invalid_encryption);
+
+    if (m_encryption.policy().stream == EncryptionPolicy::Stream::deny
+        && (m_encryption.crypto() != HandshakeEncryption::crypto_plain))
```
rakshasa, line 453:
```
if (m_encryption.policy().require_encrypted_stream() && (m_encryption.crypto() != HandshakeEncryption::crypto_rc4))
  throw handshake_error(handshake_failed, e_handshake_invalid_encryption);
```

Nothing more.

src/protocol/handshake.cc
```diff
@@ -430,32 +421,72 @@ Handshake::read_encryption_negotiation() {
...
+      else
+        error = e_handshake_invalid_encryption;
+      break;
     }
 
+    if (error)
+      throw handshake_error(message, error);
+
+    m_encryption.set_crypto(selected);
+
```
rakshasa, line 432:
```
if (m_encryption.crypto().has_stream_both()) {
  if (m_encryption.policy().preferred_stream())
    m_encryption.set_crypto(HandshakeEncryption::crypto_encrypted);
  else
    m_encryption.set_crypto(HandshakeEncryption::crypto_plaintext);

} else if (m_encryption.crypto().has_stream_plaintext()) {
  if (!m_encryption.policy().allow_plaintext_stream())
    throw handshake_error(handshake_failed, e_handshake_invalid_encryption);

  m_encryption.set_crypto(HandshakeEncryption::crypto_plaintext);

} else if (m_encryption.crypto().has_stream_encrypted()) {
  if (!m_encryption.policy().allow_encrypted_stream())
    throw handshake_error(handshake_failed, e_handshake_invalid_encryption);

  m_encryption.set_crypto(HandshakeEncryption::crypto_encrypted);

} else {
  throw handshake_error(handshake_failed, e_handshake_invalid_encryption);
}
```

src/protocol/handshake.cc
j-c-m: Including a stream retry is required for stream=allow and stream=prefer to work on outgoing connections in the same pattern it does for incoming, incoming is easier because we can choose directly in negotiations, but we can't choose for outgoing, we can only offer.

outgoing allow, you offer plain only stream encryption first, if that fails you retry the stream offering rc4
outgoing prefer, you offer rc4 stream first, if that fails you retry the stream offering both

src/torrent/runtime/encryption_policy.h
j-c-m: prefer for incoming would be the same as incoming allow? as we don't choose incoming.

rakshasa: deny, allow, prefer

require, prefer, allow

Think of it as grades from the direction of either plaintext or encrypted.

src/protocol/handshake.cc
rakshasa: The other side also gets a vote, and might have a preference.

As it stands, one side gets to pick its preference and we shouldn't try to override it.

PR #810, opened 2026-07-06
Add LoongArch CPU support

setarcos:


rakshasa:
Need to match e.g. `loongarch64` so it doesn't cause issues if there are new versions of the cpu.

Also update rtorrent scripts with the same changes.

PR #824, opened 2026-07-11
performance: size IDLE peer reads to the piece header

j-c-m:
This PR removes the memcpy in 99%+ of PIECE leech path, removes now un-needed sigaction() x 2.

---

Leech IDLE fills sizeof_piece when the download buffer is clean and the
request pipe is non-empty, so PIECE payload stays on READ_PIECE → mmap.
Incomplete messages and seed/metadata connections fill the full
ProtocolBuffer (buffer_size) to cut syscalls on control traffic. Remove
unused PeerConnectionBase::read_size.

On the clean path the body is recv'd on READ_PIECE straight into mmap,
which almost completely removes the residual from_buffer memcpy that
followed a 64-byte fill (~50 B of each 16 KiB block).

### DIFF
--- a/src/protocol/peer_connection_leech.cc
+++ b/src/protocol/peer_connection_leech.cc
@@ -349,36 +349,63 @@ PeerConnection<type>::event_read() {
 
     // Normal read.
     //
-    // We rarely will read zero bytes as the read of 64 bytes will
-    // almost always either not fill up or it will require additional
-    // reads.
+    // Leech clean IDLE fills sizeof_piece so PIECE payload stays on
+    // READ_PIECE → mmap; incomplete messages top up to buffer_size.
+    // Seed / initial-seed never receive PIECE; fill the full protocol
+    // buffer to reduce syscalls on REQUEST/HAVE control traffic.
     //
-    // Only loop when end hits 64.
+    // Loop when the fill budget is met (socket may still have data) or
+    // an incomplete message still needs more bytes.
 
     do {
 
       switch (m_down->get_state()) {
-      case ProtocolRead::IDLE:
-        if (m_down->buffer()->size_end() < read_size) {
-          unsigned int length = read_stream_throws(m_down->buffer()->end(), read_size - m_down->buffer()->size_end());
+      case ProtocolRead::IDLE: {
+        const uint32_t target = [&] {
+          // Leech clean with in-flight requests: piece header only; else full buffer.
+          if (type != Download::CONNECTION_LEECH)
+            return ProtocolRead::buffer_size;
+
+          if (m_down->buffer()->remaining() == 0 && request_list()->pipe_size() != 0)
+            return ProtocolRead::sizeof_piece;
+
+          return ProtocolRead::buffer_size;
+        }();
+
+        if (m_down->buffer()->size_end() < target) {
+          unsigned int length = read_stream_throws(m_down->buffer()->end(), target - m_down->buffer()->size_end());
           m_down->throttle()->node_used_unthrottled(length);
+
+          if (length == 0) {
+            m_down->buffer()->move_unused();
+            return;
+          }
 
           if (is_encrypted())
             m_encryption.decrypt(m_down->buffer()->end(), length);
 
           m_down->buffer()->move_end(length);
+        } else {
+          // At fill target with no room left but still an incomplete message
+          // (corrupt length / hostile peer). Without this, remaining > 0 would
+          // loop forever with no further reads.
+          if (m_down->buffer()->remaining() > 0)
+            throw communication_error("PeerConnection::event_read() download buffer full with incomplete message.");
         }
 
         while (read_message())
           ; // Do nothing.
 
-        if (m_down->buffer()->size_end() == read_size) {
+        // Continue for another fill when we hit the budget, or when an
+        // incomplete message still needs more bytes (poll-churn fix).
+        if (m_down->buffer()->remaining() > 0 || m_down->buffer()->size_end() == target) {
           m_down->buffer()->move_unused();
           break;
         } else {
           m_down->buffer()->move_unused();
           return;
         }
+      }
 
       case ProtocolRead::READ_PIECE:
         if (type != Download::CONNECTION_LEECH)
--- a/src/protocol/peer_connection_metadata.cc
+++ b/src/protocol/peer_connection_metadata.cc
@@ -188,19 +188,14 @@ PeerConnectionMetadata::event_read() {
 
   try {
 
-    // Normal read.
-    //
-    // We rarely will read zero bytes as the read of 64 bytes will
-    // almost always either not fill up or it will require additional
-    // reads.
-    //
-    // Only loop when end hits 64.
+    // Normal read: fill the full protocol buffer; loop while it is full
+    // (socket may still have data).
 
     do {
       switch (m_down->get_state()) {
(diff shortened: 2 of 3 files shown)


src/protocol/peer_connection_leech.cc
```diff
@@ -349,36 +349,57 @@ PeerConnection<type>::event_read() {
...
       switch (m_down->get_state()) {
-      case ProtocolRead::IDLE:
-        if (m_down->buffer()->size_end() < read_size) {
-          unsigned int length = read_stream_throws(m_down->buffer()->end(), read_size - m_down->buffer()->size_end());
+      case ProtocolRead::IDLE: {
+        // How far to fill this iteration: piece header only when clean, else
+        // the base fill size (large enough for REQUEST/CANCEL).
+        const uint32_t target = m_down->buffer()->remaining() > 0
+                                  ? PeerConnectionBase::read_size
+                                  : read_size;
```
rakshasa: The size shouldn't be in the header with the same name.

You're changing it to sizeof_piece, then it should use that here to make it clear what the size is from.

src/data/chunk.cc
```diff
@@ -251,19 +233,13 @@ Chunk::from_buffer(const void* buffer, uint32_t position, uint32_t length) {
...
 
-        buffer = static_cast<const char*>(buffer) + data.second;
-      } while (itr.next());
-  } else {
-      throw storage_error("no space left on disk");
-  }
+    buffer = static_cast<const char*>(buffer) + data.second;
+  } while (itr.next());
 
-  sigaction(SIGBUS, &oldact, NULL);
```
rakshasa, line 265: We could optimize this using an atomic ptr, and keep a single handler.

The atomic value is set each time we do memcpy here and read of piece messages to mmap, SIGBUS handler then knows that it is a recoverable signal.

It sets another atomic value, then returns. There's no need for setjmp.

j-c-m: Yes, we need to make the area covered larger to avoid the global sigbus crash out. Because every PIECE was usually a 50byte memcpy it used to catch most of the disk space errors more gracefully than the global, but now this memcpy is basically never, we need to catch in the recv/mmap as well (it should have been there before).

j-c-m: reviewing more, I just don't think this is worth doing, the memcpy is extremely rare to never now, we already catch the recv via error.

j-c-m:
I just restored the original sigaction path, since from_buffer is rarely called vs before where it was called every PIECE, it's OK that it calls sigaction() twice, as from_buffer is virtually optimized away.

Added the EFAULT map in read for explicit storage error before it was just a generic error.

src/data/chunk.cc
rakshasa: Diskspace issues when calling recv directly to mmap'ed memory cause SIGBUS which we must handle.

The SIGBUS signal includes the address, so if we atomically store address ranges we can properly determine if the SIGBUS was due to the recv.

j-c-m:
Backed out any SIGBUS or EFAULT changes, to make this PR purely performance (avoid chunk::from_buffer on super hot leech path).

src/data/hash_queue.cc
```diff
@@ -55,6 +55,9 @@ HashQueue::push_back(ChunkHandle handle, HashQueueNode::id_type id, slot_done_ty
 
   base_type::push_back(HashQueueNode(id, hash_chunk, std::move(d)));
 
+  // Prefault/readahead before the disk thread walks the piece for SHA1.
+  base_type::back().call_willneed();
```
rakshasa: We don't want to be this aggressive with pre-loading in hash checker, as this can cause quite a lot of thrashing.

src/protocol/peer_connection_leech.cc
```diff
@@ -349,36 +349,56 @@ PeerConnection<type>::event_read() {
...
 
       switch (m_down->get_state()) {
-      case ProtocolRead::IDLE:
-        if (m_down->buffer()->size_end() < read_size) {
-          unsigned int length = read_stream_throws(m_down->buffer()->end(), read_size - m_down->buffer()->size_end());
+      case ProtocolRead::IDLE: {
+        const uint32_t target =
+          type == Download::CONNECTION_LEECH && m_down->buffer()->remaining() == 0
+            ? ProtocolRead::sizeof_piece
+            : ProtocolRead::buffer_size;
```
rakshasa: Change this to an inline lambda function so it's more readable.

src/protocol/peer_connection_leech.cc
```diff
@@ -349,36 +349,56 @@ PeerConnection<type>::event_read() {
...
           if (is_encrypted())
             m_encryption.decrypt(m_down->buffer()->end(), length);
 
           m_down->buffer()->move_end(length);
+        } else if (m_down->buffer()->remaining() > 0) {
+          // At fill target with no room left but still an incomplete message
+          // (corrupt length / hostile peer). Without this, remaining > 0 would
+          // loop forever with no further reads.
+          throw communication_error("PeerConnection::event_read() download buffer full with incomplete message.");
         }
```
rakshasa, line 394:
```
} else {
  if (remaining >0)
    throw
}
```

Makes it easier to read.

src/protocol/peer_connection_leech.cc
rakshasa: Can't we also check if request queue is not empty?

j-c-m: Still getting a grasp on lots of the code, I think it makes sense, I used pipe_size in the update, which I think should handle control traffic slightly better. The little 13 byte read should show up less often.

src/protocol/peer_connection_leech.cc
```diff
@@ -349,36 +349,62 @@ PeerConnection<type>::event_read() {
...
+      case ProtocolRead::IDLE: {
+        const uint32_t target = [&] {
+          // Leech clean with in-flight requests: piece header only; else full buffer.
+          if (type == Download::CONNECTION_LEECH &&
+              m_down->buffer()->remaining() == 0 &&
+              request_list()->pipe_size() != 0)
+            return ProtocolRead::sizeof_piece;
+
+          return ProtocolRead::buffer_size;
+        }();
```
rakshasa, line 373:
```
if (type != Download::CONNECTION_LEECH)
          return ProtocolRead::buffer_size;

if (m_down->buffer()->remaining() == 0 && request_list()->pipe_size() != 0)
            return ProtocolRead::sizeof_piece;

 return ProtocolRead::buffer_size;
```

src/protocol/peer_connection_metadata.cc
```diff
@@ -188,19 +188,15 @@ PeerConnectionMetadata::event_read() {
...
+    // (socket may still have data).
 
     do {
       switch (m_down->get_state()) {
       case ProtocolRead::IDLE:
-        if (m_down->buffer()->size_end() < read_size) {
-          unsigned int length = read_stream_throws(m_down->buffer()->end(), read_size - m_down->buffer()->size_end());
+        if (m_down->buffer()->size_end() < ProtocolRead::buffer_size) {
+          unsigned int length = read_stream_throws(
+              m_down->buffer()->end(), ProtocolRead::buffer_size - m_down->buffer()->size_end());
```
rakshasa: Don't break up lines like this, we got plenty of horizontal space.

PR #829, opened 2026-07-13
fix: reconnect race after handshake drop on inactive download

j-c-m:
Small fix for rare race condition when on the usual tracker path, but evident if you are hammering on handshake code. Is also possible to trigger if an incoming connection is recieved while rehashing, blocking that peer for 600s.

### DIFF
--- a/src/protocol/handshake_manager.cc
+++ b/src/protocol/handshake_manager.cc
@@ -227,8 +227,13 @@ HandshakeManager::receive_succeeded(Handshake* ptr) {
       handshake->destroy_connection();
     };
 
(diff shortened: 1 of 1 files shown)


rakshasa:
If it is a race condition specifically when rehashing, shouldn't check if it is rehashing?

Which likely means we need a new flag in DownloadInfo.

j-c-m:
It's not specific to rehashing, any temporarily inactive torrent could trigger it.

rakshasa:
Updated https://github.com/rakshasa/libtorrent/pull/909

PR #831, opened 2026-07-14
storage: lazy open, idle FD close, and short-file resize

j-c-m:
Adds a close_idle time, files not used within this time period will be closed.

Does a lazy open avoid creating 0-byte, or priority off files.

Do not bulk-open FDs in FileList::open; prepare() opens on demand. Skip create/resize for priority-off files (restore when leaving off); allow boundary writes into off neighbors. Grow short files on write prepare so maps are not past EOF. Sync and close FDs on stop. Close idle open FDs via FileManager::close_idle (default 60s, 0 disables) on the manager tick.

### DIFF
--- a/src/download/download_main.cc
+++ b/src/download/download_main.cc
@@ -113,7 +113,7 @@ DownloadMain::open(int flags) {
     throw internal_error("Tried to open a download that is already open");
 
   // TODO: Move file_list open calls to DownloadMain.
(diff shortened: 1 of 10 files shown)


rakshasa:
This is too many different behavioral changes in one PR.

Some of them changes intentional behavior that is meant to ensure we're not dealing with externally modified underlying files.

PR #833, opened 2026-07-15
fix: grow short files on write prepare

j-c-m:
Always ftruncate to torrent size on write-capable prepare so maps are not past EOF.

### DIFF
--- a/src/torrent/data/file.cc
+++ b/src/torrent/data/file.cc
@@ -59,9 +59,20 @@ File::set_completed_chunks(uint32_t v) {
   m_completed = v;
 }
 
+// Grow on write whenever on-disk size is wrong (e.g. 0-byte stub after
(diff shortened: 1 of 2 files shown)


rakshasa:
This is one of those changes I was sceptical of.

There's really no reason why a file would be of non-zero size yet not matching the expected size. That means it was modified, corrupted during copying (etc) or the user opened a torrent with the same root dir and file path.

If we want to support something like this it should be a command to download to resize any files with wrong size and then rehash them.

j-c-m:
Yes it probably does mean something went awry on the underlying, possibly outside of rtorrent, and user driven, this just handles that rare case differently then now, where its stalled download or peer churn on storage_errors, again all very rare. Not a critical fix, so up to you.

rakshasa:
I would prefer there's a way for users to proactively fix a broken torrent download, rather than having it silently resize files.

PR #834, opened 2026-07-15
storage: close idle open FDs via FileManager

j-c-m:
This helps a lot for memory usage as the OS free's associated buffers/pages wired to the idle files that used to remain open until max files eviction.

Feature closes idle file FDs after a configurable amount of time rather than leaving them open until eviction presure. Default 60s, 0 disables.

rtorrent:

https://github.com/rakshasa/rtorrent/pull/1888

### DIFF
--- a/src/manager.cc
+++ b/src/manager.cc
@@ -110,6 +110,7 @@ Manager::receive_tick() {
 
   m_resource_manager->receive_tick();
   m_chunk_manager->periodic_sync();
+  m_file_manager->periodic_close_idle();
 
   // To ensure the downloads get equal chance over time at using
   // various limited resources, like sockets for handshakes, cycle the
--- a/src/torrent/data/file_manager.cc
+++ b/src/torrent/data/file_manager.cc
@@ -227,4 +227,30 @@ FileManager::evict_least_active_from_cache(unsigned int count) {
   return count;
 }
 
+void
+FileManager::periodic_close_idle() {
+  if (m_close_idle == 0 || empty())
+    return;
+
+  auto now  = this_thread::cached_time();
+  auto idle = std::chrono::seconds(m_close_idle);
+
+  size_type i = 0;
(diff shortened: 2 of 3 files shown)


rakshasa:
This code is rather old, so I don't exactly remember how it works, but we're already evicting stale File entries before opening iirc.

So we shouldn't need to iterate over every single entry here. And likely this part of the code needs to be refactored.

j-c-m:
It does evict stale entries currently, if it's at the limit. This is more proactive and closes the files when they are idle.

rakshasa:
You are manually iterating over all entries, while we should already have (or add) a sorted staleness queue for when evicting due to fd limits.

rakshasa:
Check out the last couple of PR merges as FM now has a last activity cache used during eviction.

This PR might be a bit too aggressive in closing fds, and the cache likely solves a lot of the issues here.

j-c-m:
This PR is pretty aggressive closing FDs. It's trying to address a different problem where if you are seeding 1000s of torrents with tens of thousands of files, most are idle and really don't need to be open.  60s could be to aggressive as a default, but other than that its been a win for me.

rakshasa:
Feels like 60s would be too aggressive also in your case, have you tested other values?

j-c-m:
I tried immediate close, which wasn't terrible but did get into some close<->open flapping, and also 10 minutes, which didn't seem a lot different than a minute. I was looking at this TODO, and just immediately was too much.

https://github.com/rakshasa/libtorrent/blob/57a348bd7cdf7824b83835b1a3a1811cbffc3c05/src/torrent/data/file_manager.h#L51

rakshasa:
I'd imagine a long idle timeout would be pretty much as effective at reducing memory usage as 60s.

Try 30min, run every 10min.

rakshasa:
Merged in new PR.

PR #835, opened 2026-07-15
storage: lazy open and skip create/resize for priority-off

j-c-m:
This feature keeps the disk cleaner by creating the files at write time vs when opening file list, and also avoids completely creating priority OFF files on disk (unless needed for completion of a non-OFF file).

FileList::open sets up paths/dirs only; prepare() opens FDs on demand. Sync and close all FDs on download stop.

Do not queue create/resize for PRIORITY_OFF; restore when leaving off and close any open FD. Allow boundary writes into off neighbors by re-queuing create/resize in create_chunk_part.

### DIFF
--- a/src/download/download_main.cc
+++ b/src/download/download_main.cc
@@ -113,7 +113,7 @@ DownloadMain::open(int flags) {
     throw internal_error("Tried to open a download that is already open");
 
   // TODO: Move file_list open calls to DownloadMain.
-  file_list()->open(true, (flags & FileList::open_no_create));
+  file_list()->open(flags & FileList::open_no_create);
 
   m_chunkList->resize(file_list()->size_chunks());
   m_chunkStatistics->initialize(file_list()->size_chunks());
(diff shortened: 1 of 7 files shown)


rakshasa:
Creating the files at open is part of minimizing corrupt states, and making any file-related errors happen at open.

This could be changed to not create priority-off files, yet flag them for creation. However we still need to download the partial chunks they are part of.

Which would be a more extensive change, with e.g. an option to save the partial chunks in session (or custom folder).

j-c-m:
I was somewhat building off this commit - https://github.com/rakshasa/libtorrent/commit/d526e249687af04d15f0589913ae53fa38022574

This PR addressed the issues of partials just by writing minimally into the adjacent, possibly off files. I think https://github.com/rakshasa/libtorrent/pull/834 probably addresses keeping a bunch of idle files open forever/long time but right now I am running both PRs.

I'll back this one out and see how that works.

PR #838, opened 2026-07-16
fix: touch each page in hard Chunk::preload walk

j-c-m:
The loop advanced first but always dereferenced data.first, so only the first page of each span was faulted, many times.


rakshasa:
https://github.com/rakshasa/libtorrent/pull/913

PR #840, opened 2026-07-16
fix: implement ChunkSelector::advance_position

j-c-m:
The cursor was never moved after finishing the piece at m_position or when refreshing priorities, so searches kept restarting from a stale index. Find the next untouched high- then normal-priority index, wrapping the bitfield, matching the old intended behavior.

### DIFF
--- a/src/download/chunk_selector.cc
+++ b/src/download/chunk_selector.cc
@@ -245,13 +245,13 @@ next_wanted_in_ranges(const Bitfield* untouched,
     return ChunkSelector::invalid_chunk;
 
   for (auto itr = ranges->find(first); itr != ranges->end() && itr->first < last; ++itr) {
-    uint32_t f = std::max(first, itr->first);
-    uint32_t l = std::min(last, itr->second);
+    uint32_t range_first = std::max(first, itr->first);
+    uint32_t range_last  = std::min(last, itr->second);
 
-    for (uint32_t i = f; i < l; ++i) {
-      if (untouched->get(i))
-        return i;
-    }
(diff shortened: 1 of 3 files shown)


src/download/chunk_selector.cc
```diff
@@ -235,18 +235,58 @@ ChunkSelector::search_linear_byte(utils::PartialQueue* pq, uint32_t index, Bitfi
...
+    return ChunkSelector::invalid_chunk;
+
+  for (auto itr = ranges->find(first); itr != ranges->end() && itr->first < last; ++itr) {
+    uint32_t f = std::max(first, itr->first);
+    uint32_t l = std::min(last, itr->second);
+
+    for (uint32_t i = f; i < l; ++i) {
+      if (untouched->get(i))
+        return i;
+    }
```
rakshasa: This should be a member function in Bitfield that searches for the first set index in a range.

src/download/chunk_selector.cc
```diff
@@ -248,10 +248,10 @@ next_wanted_in_ranges(const Bitfield* untouched,
...
     uint32_t l = std::min(last, itr->second);
 
-    for (uint32_t i = f; i < l; ++i) {
-      if (untouched->get(i))
-        return i;
-    }
+    uint32_t i = untouched->find_first_set(f, l);
+
+    if (i != l)
+      return i;
```
rakshasa: This isn't really the 70's when every byte was precious.

Please give the variables proper names.

j-c-m: lol its the C, it brings me back, not quite that far

PR #846, opened 2026-07-20
Limit tracker peer insertion to available-list capacity

acts-1631:
A tracker announce response can contain a large compact peer list. `PeerList::insert_available` reserves and processes every endpoint, bypassing `AvailableList`'s configured limit. Because duplicate checks are linear, a configured malicious tracker can cause excessive CPU work and retain far more peer entries than intended.

Reserve only the remaining available-list capacity and stop processing entries once the limit is reached.


src/torrent/peer/peer_list.cc
```diff
@@ -116,6 +122,9 @@ PeerList::insert_available(const void* al) {
   auto avail_last = m_available_list->end();
 
   for (const auto& addr : *address_list) {
+    if (m_available_list->size() >= m_available_list->max_size())
+      break;
```
rakshasa: This adds an if check for each loop, it's better to just use a while loop on first/last.

acts-1631:
Github is giving me weird caching issues so I'm going to try closing and opening an updated one.

PR #850, opened 2026-07-21
storage: last-ref unmap and defer munmap to the disk thread

j-c-m:
Unmap piece maps on last handle release and run MS_ASYNC + munmap on the
disk thread so main is not blocked on teardown.

Flush policy: dirty MAP_SHARED writeback is left mostly to the OS (kernel
page cache). Previously, last writable handles kept maps sticky on
ChunkList::m_queue until a periodic sync_chunks drain on main, which could
msync/munmap a large dirty set at once (main-thread flush storms under
fast leech). Maps are now retired as pieces complete instead of
accumulating for a bulk main-thread pass.

Last-ref unmap — drop the sticky writable hold; when refs hit zero, clear/unmap immediately (writable and RO share the same path). Previously there was a policy to have them be sticky for a bit, but this seems unnecessary now as modern OS stack here will hold these in page cache for some duration and manage the disk IO better.

Deferred teardown — MemoryChunk::unmap queues the VA to ThreadDisk, which does msync(MS_ASYNC) then munmap. This gets the blocking off the ui / peers thread, this was more of an issue on FreeBSD as msync(MS_ASYNC) eventually blocks for a long duration during a big flush storm, Linux would did see some blocking in munmap() but not as much overall. In any case, this keeps it from blocking UI and peer network transfers at all.

These changes offer a large removal opportunity as last-ref unmap means nothing new is parked on ChunkList::m_queue. Callers of the sticky sync path still exist, but for maps released this way they no longer do useful work

### DIFF
--- a/src/data/chunk_list.cc
+++ b/src/data/chunk_list.cc
@@ -162,9 +162,9 @@ ChunkList::get(size_type index, get_flags flags) {
   return ChunkHandle(node, flags & get_writable, flags & get_blocking);
 }
 
-// The chunks in 'm_queue' have been modified and need to be synced
-// when appropriate. Hopefully keeping the chunks mmap'ed for a while
-// will allow us to schedule writes at more resonable intervals.
+// Writable and read handles share the same last-ref unmap path: drop the
+// hold and munmap when references hit zero. Do not keep a sticky writable
+// ref on m_queue just to leave the chunk mmap'd for a later sync pass.
 
 void
 ChunkList::release(ChunkHandle* handle, release_flags flags) {
@@ -185,31 +185,16 @@ ChunkList::release(ChunkHandle* handle, release_flags flags) {
     handle->object()->dec_blocking();
   }
 
-  if (handle->is_writable()) {
+  if (handle->is_writable())
+    handle->object()->dec_rw();
+  else
+    handle->object()->dec_references();
 
-    if (handle->object()->writable() == 1) {
-      if (is_queued(handle->object()))
-        throw internal_error("ChunkList::release(...) tried to queue an already queued chunk.");
+  if (handle->object()->references() == 0) {
+    if (is_queued(handle->object()))
+      throw internal_error("ChunkList::release(...) tried to unmap a queued chunk.");
 
-      // Only add those that have a modification time set?
-      //
-      // Only chunks that are not already in the queue will execute
-      // this branch.
-      m_queue.push_back(handle->object());
-
-      runtime::memory_manager()->account_sync_queue(m_chunk_size);
-
-    } else {
-      handle->object()->dec_rw();
-    }
-
-  } else {
-    if (handle->object()->dec_references() == 0) {
-      if (is_queued(handle->object()))
-        throw internal_error("ChunkList::release(...) tried to unmap a queued chunk.");
-
-      clear_chunk(handle->object(), flags);
-    }
+    clear_chunk(handle->object(), flags);
   }
 
   handle->clear();
--- a/src/data/chunk_list.cc
+++ b/src/data/chunk_list.cc
@@ -205,7 +205,24 @@ ChunkList::clear_chunk(ChunkListNode* node, release_flags flags) {
   if (!node->is_valid())
     throw internal_error("ChunkList::clear_chunk(...) !node->is_valid().");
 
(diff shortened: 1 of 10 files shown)


rakshasa:
Just like the aync fd close changes I did, this should be in a separate thread.

j-c-m:
Light testing on the move to std::async, will continue more.

src/data/memory_chunk.cc
```diff
@@ -50,7 +52,20 @@ MemoryChunk::unmap() {
...
-  if (munmap(m_ptr, m_end - m_ptr) != 0)
+  void*  ptr    = m_ptr;
+  size_t length = static_cast<size_t>(m_end - m_ptr);
+
+  // Drop VA ownership on this thread; ChunkManager's MunmapQueue worker does
+  // MS_ASYNC + munmap.
+  if (manager != nullptr) {
+    manager->chunk_manager()->queue_munmap(ptr, length);
+    return;
+  }
```
rakshasa: This is the wrong place for this, MemoryChunk is supposed to be a simple type.

src/utils/munmap_queue.h
```diff
@@ -0,0 +1,44 @@
...
+#include <utility>
+#include <vector>
+#include <torrent/common.h>
+
+namespace torrent::utils {
+
+// Persistent std::async worker that issues MS_ASYNC + munmap off the caller
+// (and off ThreadDisk). Same shape as FdCloseQueue.
+
+class MunmapQueue {
```
rakshasa: `MemoryUnmapQueue`

src/utils/munmap_queue.h
```diff
@@ -0,0 +1,44 @@
...
+  MunmapQueue();
+  ~MunmapQueue();
+
+  void                queue(void* ptr, size_t length);
+
+private:
+  MunmapQueue(const MunmapQueue&) = delete;
+  MunmapQueue& operator=(const MunmapQueue&) = delete;
+
+  using region_type = std::pair<void*, size_t>;
```
rakshasa: `queue_type` with the vector

src/data/chunk_manager.h
```diff
@@ -59,6 +69,8 @@ class LIBTORRENT_EXPORT ChunkManager : private std::vector<ChunkList*> {
 
   std::chrono::seconds m_last_try_free_memory{};
   size_type            m_last_freed_index{};
+
+  std::unique_ptr<utils::MemoryUnmapQueue> m_munmap_queue;
```
rakshasa, line 73: `m_unmap_queue`

src/data/chunk_list.cc
```diff
@@ -220,7 +205,24 @@ ChunkList::clear_chunk(ChunkListNode* node, release_flags flags) {
...
+    MemoryChunk& mc = part.chunk();
+
+    if (!mc.is_valid())
+      continue;
+
+    void*  ptr    = mc.ptr();
+    size_t length = static_cast<size_t>(mc.end() - mc.ptr());
+
+    mc.clear();
+    m_manager->queue_munmap(ptr, length);
```
rakshasa, line 222: `queue_memory_unmap` and pass `Chunk*` to m_manager, and it builds a vector of ptr+length to pass to MemoryUnmapQueue.

src/data/chunk_list.cc
```diff
@@ -185,31 +185,16 @@ ChunkList::release(ChunkHandle* handle, release_flags flags) {
...
-      if (is_queued(handle->object()))
-        throw internal_error("ChunkList::release(...) tried to queue an already queued chunk.");
-
-      // Only add those that have a modification time set?
-      //
-      // Only chunks that are not already in the queue will execute
-      // this branch.
-      m_queue.push_back(handle->object());
-
-      runtime::memory_manager()->account_sync_queue(m_chunk_size);
```
rakshasa, line 200: You are no longer accounting for sync queue in memory_manager.

This is also removing `m_queue` usage completely, which has a lot of issues, see `ChunkList::sync_chunks`

You can only use the async unmap queue when it is safe, e.g. `!runtime::memory_manager()->safe_sync()` and not low diskspace.

rakshasa (commented):
Forgot to submit review.

PR #851, opened 2026-07-23
Limit tracker peer insertion to available-list capacity

acts-1631:
A tracker announce response can contain a large compact peer list. `PeerList::insert_available` reserves and processes every endpoint, bypassing `AvailableList`'s configured limit. Because duplicate checks are linear, a configured malicious tracker can cause excessive CPU work and retain far more peer entries than intended.

Reserve only the remaining available-list capacity and stop processing entries once the limit is reached.

### DIFF
--- a/src/torrent/peer/peer_list.cc
+++ b/src/torrent/peer/peer_list.cc
@@ -101,13 +101,19 @@ uint32_t
 PeerList::insert_available(const void* al) {
   auto address_list = static_cast<const AddressList*>(al);
 
+  if (m_available_list->size() >= m_available_list->max_size())
+    return 0;
+
   uint32_t inserted = 0;
   uint32_t invalid = 0;
   uint32_t unneeded = 0;
   uint32_t updated = 0;
 
-  if (m_available_list->size() + address_list->size() > m_available_list->capacity())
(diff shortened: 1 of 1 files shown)


rakshasa:
Forgot how complex the for loop was.

You should only ever process a max number of entries, even if they get rejected or are duplicates. Otherwise the tracker can just feed us more and more entries each time.

Probably need to review this.

acts-1631:
Would bounding the input range up front be what you had in mind?

```c++
auto remaining = m_available_list->max_size() - m_available_list->size();
auto addr_itr = address_list->begin();
auto addr_last = addr_itr + std::min(remaining, address_list->size());

while (addr_itr != addr_last) {
  const auto& addr = *addr_itr++;
  // Existing validation, lookup, and insertion.
}
```

This limits inspected entries as well as inserted entries, so invalid or duplicate tracker entries cannot extend processing past the available capacity.

rakshasa:
I'll need to review all the related code, the address_list passed needs checked before passing to this function, etc.

Although not sure if we can just ignore new entries as that might mean an m_available_list that is full of dead / bad peers might cause us to stall as we get no more fresh peers.

PR #852, opened 2026-07-25
net: do not throw when a start callback outlives its CurlGet

xirvik:
CurlGet::start() queues start_get() on the stack thread, and close_self() on a get that was never stacked resets the object so it can be reused. Thread::cancel_callback() cannot stop a callback that has already been dispatched, so start_get() could run against a reset get and throw

  CurlGet::prepare_start(...) called on an object that was not started.

from the net thread. Thread::event_loop() rethrows internal_error, which terminates the process.

Seen in production with a tracker announce that is closed and re-armed in the same instant, e.g. when a stop event is followed immediately by a start event for the same download: the stopped announce is closed before the net thread has run its start callback, and the process aborts minutes later.

Track the pending start explicitly so a withdrawn start becomes a no-op rather than a fatal error. The added test reproduces the sequence and fails with the above exception without the fix.

### DIFF
--- a/src/net/curl_get.cc
+++ b/src/net/curl_get.cc
@@ -179,8 +179,9 @@ CurlGet::start(const std::shared_ptr<CurlGet>& curl_get, CurlStack* stack) {
   if (self->m_was_started)
     throw torrent::internal_error("CurlGet::start() called on an already started object.");
 
-  self->m_was_started  = true;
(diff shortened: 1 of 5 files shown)


rakshasa:
Seems we should just use shared_ptr's created per-request, and the callbacks should check a weak_ptr.

This would simplify a lot of the logic.

PR #853, opened 2026-07-25
Avoid blocking on fd close using a separate thread by @j-c-m

rakshasa:


rakshasa:
This is still inefficient as it only evicts one file in most cases, it should use max open files to decide on a number to evict each time.

Or build a cache of entries.

PR #855, opened 2026-07-25
fix: FileManager::detach least active cache fix

j-c-m:
Capture the closed File* before swap-remove and erase that entry.


src/torrent/data/file_manager.cc
```diff
@@ -108,16 +108,17 @@ FileManager::detach(File* file) {
 
 int
 FileManager::detach(iterator itr) {
-  int fd = (*itr)->file_descriptor();
+  File* closed = *itr;
+  int   fd     = closed->file_descriptor();
```
rakshasa: `auto* file = *itr;`

PR #860, opened 2026-07-28
fix: FileManager check if the file can be closed

jmakovicka:
Ignore files that are not open, or are padding. In case some cached files are already closed, evict_least_active will receive a larger number of closed files than actually closed, resulting in "failed to evict enough files".

### DIFF
--- a/src/torrent/data/file_manager.cc
+++ b/src/torrent/data/file_manager.cc
@@ -232,6 +232,11 @@ FileManager::evict_least_active_from_cache(unsigned int count) {
     if (pair.first->last_touched() != pair.second)
       continue;
 
+    if (!pair.first->is_open())
+      continue;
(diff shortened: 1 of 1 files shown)


src/torrent/data/file_manager.cc
```diff
@@ -232,6 +232,9 @@ FileManager::evict_least_active_from_cache(unsigned int count) {
     if (pair.first->last_touched() != pair.second)
       continue;
 
+    if (!pair.first->is_open() || pair.first->is_padding())
+      continue;
```
rakshasa, line 236: The cache shouldn't contain any `is_padding()` files.

rakshasa:
Note for future PRs that we now have a linter running on commit message sizes.

PR #861, opened 2026-07-29
Bound HTTP tracker scrape responses

acts-1631:
`TrackerHttp::send_event_unsafe()` limits announce responses to 1 MiB
and restricts redirects to HTTP/HTTPS before starting the request.
`send_scrape_unsafe()` also resets the `HttpGet`, but did not restore
either setting. Since `HttpGet::reset()` creates a fresh `CurlGet`, an
attacker-controlled tracker could return an oversized scrape response
that would be buffered in the scrape `stringstream`.

This applies the existing announce response limit and redirect policy to
scrape requests as well.

### DIFF
--- a/src/tracker/tracker_http.cc
+++ b/src/tracker/tracker_http.cc
@@ -199,6 +199,9 @@ TrackerHttp::send_scrape_unsafe() {
   m_get.reset(request_url, m_data);
   m_get.use_family(m_current_family);
 
+  m_get.set_max_file_size(1 << 20);
+  m_get.set_redirect_only_http_https();
+
   m_get.add_done_slot(tracker_thread::thread(), [this] { receive_done(); });
   m_get.add_failed_slot(tracker_thread::thread(), [this](const auto& str) { receive_signal_failed(str); });
 


src/net/curl_get.cc
```diff
@@ -496,10 +498,21 @@ CurlGet::trigger_cleared_request() {
...
+  if (size != 0 && nmemb > std::numeric_limits<size_t>::max() / size)
     return 0;
 
-  return size * nmemb;
+  size_t length = size * nmemb;
+
+  if (handle->m_max_file_size != 0 &&
+      (handle->m_size_received > handle->m_max_file_size ||
+       length > handle->m_max_file_size - handle->m_size_received))
+    return 0;
```
rakshasa: This seems unnecessary if libcurl already fixed the issue and it isn't a critical vulnerability.

acts-1631: That makes sense. The callback check was only intended to cover older libcurl versions, since `configure.ac` does not set a minimum version. I'll drop that part.

rakshasa: It's more about code complexity, it's a bigger risk to add this much code than whatever vulnerability the libcurl issue introduced. (which is very limited)

Since it's fixed now it's better to just keep things clean.

acts-1631: Is there anything else you want done with this one or is it ok to go?

PR #872, opened 2026-08-07
runtime: fix the argument count in the listen restart error log.

xirvik:
The line has two conversions and one argument, so %s reads a garbage vararg.

### DIFF
--- a/src/torrent/runtime/network_manager.cc
+++ b/src/torrent/runtime/network_manager.cc
@@ -264,7 +264,7 @@ NetworkManager::listen_restart_unsafe() {
     listen_open_unsafe(m_listen_port, m_listen_port);
 
   } catch (const base_error& e) {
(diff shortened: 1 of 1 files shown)


src/torrent/runtime/network_manager.cc
```diff
@@ -264,7 +264,7 @@ NetworkManager::listen_restart_unsafe() {
     listen_open_unsafe(m_listen_port, m_listen_port);
 
   } catch (const base_error& e) {
-    LT_LOG_NOTICE("Could not restart listen socket: %" PRIu16 " : %s", e.what());
+    LT_LOG_NOTICE("Could not restart listen socket: %s", e.what());
```
rakshasa: This should instead add `m_listen_port`

xirvik: > This should instead add `m_listen_port`

Done

PR #874, opened 2026-08-08
Do not allocate a bencode string before reading it.

xirvik:
TL;DR The declared length comes from the input, so a 26 byte reply could reserve 4 GB.

object_read_string() sizes the string from the length written in the input, before reading any of it:

  uint32_t size;
  *input >> size;
  ...
  str.resize(size);

  resize() allocates and zero-fills, so a reply that declares a large string and then delivers nothing has already cost the memory. The std::length_error catch around it only fires on a 32-bit build, where the request exceeds max_size(); on 64-bit the allocation is legal and simply
  succeeds.

  A 26 byte HTTP tracker reply is enough:

  d10:xxxxxxxxxx4294967295:e

  The change reads the string in 8 KB blocks and appends what actually arrives, so memory tracks the data instead of the claim. It also replaces the per-character istream::get() loop, so the common path is no slower — loading a torrent with a 2 MB pieces string measured 46 ms before and 45 ms after.

  object_read_bencode_c_string(), the char-buffer parser used for DHT packets, already bounds the declared length against the bytes available; this brings the istream path in line with it.

### DIFF
--- a/src/torrent/object_stream.cc
+++ b/src/torrent/object_stream.cc
@@ -1,5 +1,7 @@
 #include "config.h"
 
+#include <algorithm>
+
 #include "torrent/object_stream.h"
 
 #include <iostream>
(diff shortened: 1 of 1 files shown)


rakshasa:
Merged.

PR #897, opened 2026-08-17
Don't pass a negative char to the ctype functions.

xirvik:
A peer id is arbitrary bytes, isalpha is defined for unsigned char values.

(pure portability; no observable issues)

### DIFF
--- a/src/torrent/peer/client_list.cc
+++ b/src/torrent/peer/client_list.cc
@@ -198,8 +198,8 @@ ClientList::insert_helper(ClientInfo::id_type type,
 bool
 ClientList::retrieve_id(ClientInfo* dest, const HashString& id) const {
   if (id[0] == '-' && id[7] == '-' &&
-      std::isalpha(id[1]) && std::isalpha(id[2]) &&
-      std::isxdigit(id[3]) && std::isxdigit(id[4]) && std::isxdigit(id[5]) && std::isxdigit(id[6])) {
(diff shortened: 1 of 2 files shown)


src/torrent/peer/client_list.cc
```diff
@@ -194,12 +195,21 @@ ClientList::insert_helper(ClientInfo::id_type type,
...
+id_is_xdigit(char c) { return std::isxdigit(static_cast<unsigned char>(c)); }
+
 // Make this properly honor const-ness.
 bool
 ClientList::retrieve_id(ClientInfo* dest, const HashString& id) const {
   if (id[0] == '-' && id[7] == '-' &&
-      std::isalpha(id[1]) && std::isalpha(id[2]) &&
-      std::isxdigit(id[3]) && std::isxdigit(id[4]) && std::isxdigit(id[5]) && std::isxdigit(id[6])) {
+      id_is_alpha(id[1]) && id_is_alpha(id[2]) &&
+      id_is_xdigit(id[3]) && id_is_xdigit(id[4]) && id_is_xdigit(id[5]) && id_is_xdigit(id[6])) {
```
rakshasa: These changes feel wrong, we should have a way to retrieve the unsigned char type when calling std::is*

xirvik: > These changes feel wrong, we should have a way to retrieve the unsigned char type when calling std::is*

Fixed with at_uchar

PR #898, opened 2026-08-17
Let the dns class enum cover the whole wire range.

xirvik:
TLDR; A class is 16 bits on the wire, and dns_type already carries DNS_T_MAX.

 A resource record class is a 16 bit field on the wire, but the enumerators in enum dns_class stop at DNS_C_ANY = 255, so dns_initparse() and dns_nextrr() cast  a larger value into an enum whose range does not include it. enum dns_type  already carries DNS_T_MAX = 65536 for exactly this reason; this adds the  equivalent for classes.


rakshasa:
This is a pointless change, and if it does something it should be pushed upstream to udns.

PR #899, opened 2026-08-20
Detect overflow when reading a bencode integer.

xirvik:
The accumulator wrapped, so i9223372036854775808e parsed as the smallest integer.


rakshasa:
https://github.com/rakshasa/libtorrent/pull/903

PR #900, opened 2026-08-20
Reject a bencode string length that does not fit in unsigned int.

xirvik:
The accumulator wrapped, so 4294967300: was accepted as a four byte string.


rakshasa:
https://github.com/rakshasa/libtorrent/pull/902

PR #904, opened 2026-08-24
Reject leading zeros in a bencode integer regardless of sign.

xirvik:
The check read the parsed value, so i00e and i-01e were accepted.

### DIFF
--- a/src/torrent/object_stream.cc
+++ b/src/torrent/object_stream.cc
@@ -59,11 +59,17 @@ object_read_bencode_c_value(const char* first, const char* last, int64_t& value)
   if (errc.ptr >= last || *errc.ptr != 'e')
     throw torrent::bencode_error("Invalid bencode data: missing 'e' terminator.");
 
-  if (value != 0 && *first == '0')
-    throw torrent::bencode_error("Invalid bencode data: leading zeros are not allowed.");
+  if (*first == '-') {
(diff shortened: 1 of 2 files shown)


src/torrent/object_stream.cc
rakshasa: Avoid `?` when possible as it is hard to read, and another unnessesary char* adds complexity.

This whole thing can be restructured by an `if (*first == '-') { do checks } else { do checks }`

src/torrent/object_stream.cc
```diff
@@ -59,11 +59,17 @@ object_read_bencode_c_value(const char* first, const char* last, int64_t& value)
...
+    if (value == 0)
+      throw torrent::bencode_error("Invalid bencode data: negative zero is not allowed.");
 
-  if (value == 0 && *first == '-')
-    throw torrent::bencode_error("Invalid bencode data: negative zero is not allowed.");
+    if (errc.ptr - first > 2 && *(first + 1) == '0')
+      throw torrent::bencode_error("Invalid bencode data: leading zeros are not allowed.");
+
+  } else {
+    if (errc.ptr - first > 1 && *first == '0')
```
rakshasa:
```
if (*(first + 1) == '0' && errc.ptr != first + 2)

if (*first == '0' && errc.ptr != first + 1)
```

PR #906, opened 2026-08-26
Add DHT peer nodes learned from the peer protocol PORT message.

xirvik:
NetworkManager::dht_add_peer_node has an empty body — the arguments are [[maybe_unused]] and the comment says peer nodes are ignored, with a note to re-enable if it causes issues. So every DHT node advertised over the peer protocol's PORT message is discarded, and the DHT router only ever learns nodes from bootstrap and from DHT traffic itself.
  
Pass them to m_dht_controller->add_node(sa, port), which is what the callers already assume happens.

### DIFF
--- a/src/dht/dht_router.cc
+++ b/src/dht/dht_router.cc
@@ -22,8 +22,9 @@ namespace torrent {
 
 HashString DhtRouter::zero_id;
 
-DhtRouter::DhtRouter(const Object& cache)
+DhtRouter::DhtRouter(tracker::DhtController* controller, const Object& cache)
   : DhtNode(zero_id, sa_make_inet_any().get()), // actual ID is set later
+    m_controller(controller),
     m_server(this),
     m_curToken(random()),
     m_prevToken(random()),
@@ -106,6 +107,8 @@ DhtRouter::start(int port) {
   LT_LOG_THIS("starting: port:%d", port);
 
   m_server.start(port);
+
+  m_controller->set_nodes_populated(check_nodes_populated());
 
   // Set timeout slot and schedule it to be called immediately for initial bootstrapping if
   // necessary.
@@ -403,7 +406,9 @@ DhtRouter::receive_timeout_bootstrap() {
   // we have enough nodes in our routing table. After we have 32 nodes, we switch
   // to a less aggressive non-bootstrap mode of collecting nodes that contact us
   // and through doing normal torrent announces.
-  if (m_nodes.size() < num_bootstrap_complete) {
+  m_controller->set_nodes_populated(check_nodes_populated());
+
+  if (!check_nodes_populated()) {
     if (!m_contacts.has_value())
       throw internal_error("DhtRouter::receive_timeout_bootstrap called without contact list.");
 
@@ -480,6 +485,8 @@ DhtRouter::receive_timeout() {
     }
     ++itr;
   }
+
+  m_controller->set_nodes_populated(check_nodes_populated());
 
   m_server.update();
 
--- a/src/torrent/runtime/network_manager.cc
+++ b/src/torrent/runtime/network_manager.cc
@@ -169,10 +169,11 @@ NetworkManager::dht_add_bootstrap_node(std::string host, int port){
 }
 
 void
-NetworkManager::dht_add_peer_node([[maybe_unused]] const sockaddr* sa, [[maybe_unused]] int port) {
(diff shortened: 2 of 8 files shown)


src/torrent/runtime/network_manager.cc
```diff
@@ -169,10 +169,8 @@ NetworkManager::dht_add_bootstrap_node(std::string host, int port){
 }
 
 void
-NetworkManager::dht_add_peer_node([[maybe_unused]] const sockaddr* sa, [[maybe_unused]] int port) {
-  // Ignore peer nodes as we shouldn't need them(?)
-  //
-  // Re-enable this if it causes issues.
+NetworkManager::dht_add_peer_node(const sockaddr* sa, int port) {
+  m_dht_controller->add_node(sa, port);
```
rakshasa, line 176: This really needs to have an atomic bool check in DhtController to see if we need more peers.

src/dht/dht_router.cc
```diff
@@ -481,6 +490,8 @@ DhtRouter::receive_timeout() {
     ++itr;
   }
 
+  m_controller->set_wanting_nodes(m_nodes.size() < num_bootstrap_complete);
```
rakshasa: I've seen this `m_nodes.size() < num_bootstrap_complete` too many times, it should be an inline method.

src/dht/dht_router.cc
```diff
@@ -403,7 +408,11 @@ DhtRouter::receive_timeout_bootstrap() {
   // we have enough nodes in our routing table. After we have 32 nodes, we switch
   // to a less aggressive non-bootstrap mode of collecting nodes that contact us
   // and through doing normal torrent announces.
-  if (m_nodes.size() < num_bootstrap_complete) {
+  bool wanting_nodes = m_nodes.size() < num_bootstrap_complete;
+
+  m_controller->set_wanting_nodes(wanting_nodes);
+
+  if (wanting_nodes) {
```
rakshasa: If you're adding this many lines of code to optimize a bool check you're doing something wrong.

E.g. set_wanting_node should be inline and comparing it twice will be optimized away as long as you use the correct atomic store.

src/torrent/tracker/dht_controller.cc
```diff
@@ -40,6 +40,11 @@ DhtController::is_receiving_requests() {
   return m_receive_requests;
 }
 
+bool
+DhtController::is_wanting_nodes() {
+  return m_wanting_nodes.load(std::memory_order_relaxed);
+}
```
rakshasa: This should be inlined.

When inlining, add it as a separate function after the class declaration in the header.

src/torrent/tracker/dht_controller.cc
```diff
@@ -108,6 +113,11 @@ DhtController::set_receive_requests(bool state) {
   m_receive_requests = state;
 }
 
+void
+DhtController::set_wanting_nodes(bool state) {
+  m_wanting_nodes.store(state, std::memory_order_relaxed);
+}
+
```
rakshasa: Same, inlined.

src/torrent/runtime/network_manager.cc
```diff
@@ -169,10 +169,11 @@ NetworkManager::dht_add_bootstrap_node(std::string host, int port){
 }
 
 void
-NetworkManager::dht_add_peer_node([[maybe_unused]] const sockaddr* sa, [[maybe_unused]] int port) {
-  // Ignore peer nodes as we shouldn't need them(?)
-  //
-  // Re-enable this if it causes issues.
+NetworkManager::dht_add_peer_node(const sockaddr* sa, int port) {
+  if (!m_dht_controller->is_wanting_nodes())
```
rakshasa: Kinda don't like the `wanting` verb here, feels wrong.

The function should instead use `is_*` to say it has enough nodes, though blanking out on a good name.

rakshasa: `is_nodes_populated` or full, or somethings

src/dht/dht_router.cc
```diff
@@ -481,6 +486,8 @@ DhtRouter::receive_timeout() {
     ++itr;
   }
 
+  m_controller->set_nodes_populated(is_nodes_populated());
```
rakshasa: You can't name them both like that, at least make it `check_nodes_populated()`

PR #919, opened 2026-09-08
Size the request pipe from a recent rate counter, not a 30 s average

silasmariusz:
PeerConnectionBase::try_request_pieces() sizes a peer's outstanding request queue from that peer's measured download rate, and Rate::rate() divides by the counter's whole span rather than by the time it has actually been collecting:

    src/torrent/rate.cc
        return m_current / m_span;

The peer download counter spans thirty seconds (peer_chunks.h, ThrottleNode m_downloadThrottle{30}), so one second into a connection it reports a thirtieth of what the peer is really sending, a sixth after five seconds, and is only right after thirty.

The request pipe is what that lands on. At rate zero it is two blocks -- test/protocol/test_request_list.cc already asserts calculate_pipe_size(0) == 2 -- which is 32 KiB in flight, a ceiling of 32 KiB per round trip: roughly 1.6 MB/s to a peer 20 ms away, roughly 210 KB/s to one 150 ms away. It widens as the counter catches up, and the span sets how fast: each round trip delivers 16 KiB per outstanding block, lifting the reading by 16/span KB/s per block, so the pipe grows by about 1 + 3.2/span per round trip above 20 KB/s. At span 30 that is some 11% per round trip, which is seconds of ramp on a distant peer.

Add a second counter to ThrottleNode with a five-second span, fed from the single place the existing one is fed (ThrottleList::node_used), and size only the request pipe from it. The thirty-second counter, calculate_pipe_size() and its formula, the throttle scheduler and every reported rate are unchanged.

The counter lives in ThrottleNode rather than in PeerChunks because node_used() is the one point every payload path reaches -- peer_connection_base.cc alone calls it from three places -- so it cannot silently miss a path added later. The cost is one std::deque of at most five entries per node.

Five seconds rather than one or two because Rate buckets by whole seconds, so a very short span is one or two buckets and reads as a step function; try_request_pieces() already damps the rest by only topping the pipe up once the depth falls below (pipeSize + 10) / 2.


rakshasa:
This should be solved without adding a new `Rate` object for every peer.

Might as well reduce the peer's rate span.

Issue #920, opened 2026-09-09
Do not create `PRIORITY_OFF` files; keep the partial chunks in a per-download sidecar


noctuum:
> This could be changed to not create priority-off files, yet flag them for creation. However we still need to download the partial chunks they are part of. Which would be a more extensive change, with e.g. an option to save the partial chunks in session (or custom folder).
>
> Originally posted by @rakshasa in https://github.com/rakshasa/libtorrent/pull/835#issuecomment-5019423215

Opening this to settle the design before writing any code, rather than putting a large change in front of you unannounced.

Prior art, all of it still open unless noted:

- **PR #835** — "storage: lazy open and skip create/resize for priority-off", by @j-c-m. The direct predecessor; this issue is about the part its review left open.
- **PR #753** (merged) — "fix: skip fallocate for priority=0 files in resize_file()", by @trim21.
- **rakshasa/rtorrent#19** (issue, 2012) — "Enhancement: Hide the files that aren't to be downloaded", where you called this a planned feature.
- **rakshasa/rtorrent#143** (issue, 2013) — "Files set as 'Do Not Download' are still created and may contain data".
- **rakshasa/rtorrent#588** (issue, 2017) — "feature request: improve partial file support", which already describes the oversized-neighbour symptom and names the two chunk classes: *ignored*, needed only by don't-download files, and *partial*, shared with a wanted file. I use those terms below.

## Goal

No `PRIORITY_OFF` file is created on disk — including one adjacent to a wanted file.

Today every skipped file is created as a zero-length stub at `DownloadMain::start`, and a skipped file sharing a partial chunk with a wanted one is additionally `ftruncate`d to its full length the first time that chunk is written, since `resize_file()` runs from `File::prepare` once the file is opened with write permission. So selecting one file out of a directory of 40 GB files leaves a zero-length stub for each skipped file, plus a 40 GB sparse file on each side of the selection holding a single chunk of real data. `fallocate` is correctly skipped for `PRIORITY_OFF` (#753), so the blocks are not allocated — but the apparent size is, which is enough to confuse `du --apparent-size`, backup tools, `rsync` without `--sparse`, and any move to a filesystem without sparse support.

Deleting those two stubs is not an option either: the partial chunk stops verifying, the wanted file drops below 100%, the chunk is re-downloaded, and the stub comes back.

[...]

PR #926, opened 2026-09-11
Initialize and clear MemoryChunk's protection flags, drop the unused m_flags

noctuum:
## Problem

- `m_prot` had no initialiser, so a default-constructed chunk carried an indeterminate protection mask. `is_readable()`, `is_writable()`, `is_exec()`, `has_permissions()` and `get_prot()` all read it.
- `clear()` reset every member except `m_prot`, so a cleared chunk kept the protection of the mapping it used to have.
- `m_flags` was never read anywhere: set in the constructor, reset in `clear()`, and nothing else.

## Fix

- `m_prot{PROT_NONE}`, and `clear()` resets it. A chunk with no mapping now grants nothing.
- `m_flags` removed, along with the constructor argument that fed it. Both `SocketFile` call sites were passing the flags they had already given to `mmap`, so the mapping is unaffected.

## Test

Verified by building and running `make check`, which is unaffected.

Fifth of the cleanup PRs listed in #920.


src/data/memory_chunk.h
```diff
@@ -92,7 +92,7 @@ class MemoryChunk {
   char*               m_begin{};
   char*               m_end{};
 
-  int                 m_prot;
+  int                 m_prot{PROT_NONE};
   int                 m_flags{PROT_NONE};
```
rakshasa: It looks like we don't use `m_flags` anywhere. Also `clear()` doesn't clear `m_prot`.

noctuum: Both done. `m_flags` and the constructor argument that fed it are gone; the two `SocketFile` call sites were handing over the same flags they had already passed to `mmap`. `clear()` now resets `m_prot`.

PR #939, opened 2026-09-16
Fix the last metadata piece length

noctuum:
The last piece length is computed as `metadata_size % 16384`, which is 0 when the metadata size is a multiple of 16384. The sender transmits an empty last piece. A leeching rtorrent 0.16.22 rejected it and asked again 1130 times in eight seconds, and the magnet never completed.

A 32768 byte info dict is an ordinary torrent:

| piece length | torrent size |
|---|---|
| 256 KB | 407 MB |
| 1 MB | 1.6 GB |

Both metadata pieces requested from a seeding rtorrent 0.16.22:

| piece | before | after |
|---|---|---|
| 0 | 16384 | 16384 |
| 1 | 0 | 16384 |


src/protocol/peer_connection_metadata.cc
```diff
@@ -420,7 +420,7 @@ PeerConnectionMetadata::receive_metadata_piece(uint32_t piece, const char* data,
     length = ProtocolExtension::metadata_piece_size;
 
     if ((piece << ProtocolExtension::metadata_piece_shift) + ProtocolExtension::metadata_piece_size >= m_download->file_list()->size_bytes())
-      length = m_download->file_list()->chunk_size() % ProtocolExtension::metadata_piece_size;
+      length = m_download->file_list()->size_bytes() - (piece << ProtocolExtension::metadata_piece_shift);
```
rakshasa: These two should call a helper function to get the length, with proper if statements.

noctuum: Done - both now call `ProtocolExtension::metadata_piece_length(piece, metadata_size)`, out of line in `extensions.cc` with the ifs spelled out.

PR #961, opened 2026-09-19
Fix one-byte overflow of the domain name buffer in udns dns_getdn

xirvik:
The terminator guard tested the buffer base rather than the write cursor.


rakshasa:
This seems like it should also be reported upstream.

xirvik:
> This seems like it should also be reported upstream.

Yes, I sent 5 udns patches today.

PPN-SD:
> > This seems like it should also be reported upstream.
> 
> Yes, I sent 5 udns patches today.

Hi, just out of curiosity, did upstream ever get back to you about the patches, including Rakshasa’s older patches?

Here (Gentoo) generally [try to use external libs whenever possible](https://devmanual.gentoo.org/general-concepts/bundled-deps/index.htm), but since udns is modified (and I see other PR), I don't think unbundling it is really an option anymore.

xirvik:
> > > This seems like it should also be reported upstream.
> > 
> > 
> > Yes, I sent 5 udns patches today.
> 
> Hi, just out of curiosity, did upstream ever get back to you about the patches, including Rakshasa’s older patches?
> 
Yes, within hours.

PR #963, opened 2026-09-19
Limit the decoded response size in CurlGet

xirvik:
CURLOPT_MAXFILESIZE counts wire bytes, so a compressed body evaded the cap.


rakshasa:
libcurl 8.2.0 already fixed this, so not needed.

xirvik:
> libcurl 8.2.0 already fixed this, so not needed.

Yes, but libtorrent doesn't have a minimum version specified in configure.ac, and the current version in Debian stable is 8.14.1.

rakshasa:
I wouldn't remove it if it was already in the code, however this is adding extra code for something that I've never had any bug reports about and the upstream fix will be used by almost everyone soon anyway.

xirvik:
> I wouldn't remove it if it was already in the code, however this is adding extra code for something that I've never had any bug reports about and the upstream fix will be used by almost everyone soon anyway.

Most things we're fixing don't have a bug report yet, but this is exploitable and these days anyone can do that ...  I won't push for this if you really prefer this PR to be abandoned, but it's a real thing.

PR #967, opened 2026-09-19
Keep PeerInfo::id_hex NUL-terminated

xirvik:
The handshake writes 40 hex characters, so %40s read past the array into the next member.


rakshasa:
This is the wrong solution, instead fix the printf statements by using `%40.40s`

rakshasa:
Those tests are not needed and just adds bloat to the codebase.

xirvik:
> Those tests are not needed and just adds bloat to the codebase.

Removed

rakshasa:
For future PR's, instruct the LLMs to not generate unit tests that have marginal utility as unit tests make code harder to change.

xirvik:
> For future PR's, instruct the LLMs to not generate unit tests that have marginal utility as unit tests make code harder to change.

TBH I consider unit tests quite critical these days (the harder to change being a feature) even though I admit these ones are not very useful. 
I'll do it for libtorrent/rtorrent anyway.

PR #968, opened 2026-09-19
Fix the signal mask and the reported cause on a chunk write bus error

xirvik:
Use sigsetjmp, jump only when the thread armed it, and stop blaming disk space.


src/data/chunk.cc
```diff
@@ -48,12 +48,23 @@
...
+bus_handler(int signum, siginfo_t*, void*) {
+  if (!jmp_disk_full_armed) {
+    struct sigaction sa{};
+
+    sa.sa_handler = SIG_DFL;
+    sigemptyset(&sa.sa_mask);
+
+    sigaction(signum, &sa, nullptr);
+    return;
+  }
```
rakshasa, line 66: `SignalHandler::set_sigaction_handler(SIGBUS, &handle_sigbus);`

You are removing this action.

xirvik: > 
> You are removing this action.

Thanks - fixed. Now when the guard is not armed, bus_handler() reinstalls the disposition saved from before it was installed.

PR #969, opened 2026-09-19
Reject torrents with more chunks than the chunk index can address

xirvik:
The count is narrowed to the uint32_t bitfield size and can truncate to zero.


src/torrent/data/file_list.cc
```diff
@@ -349,6 +349,9 @@ FileList::initialize(uint64_t torrentSize, uint32_t chunkSize) {
   if (chunkSize == 0)
     throw internal_error("FileList::initialize() chunk_size() == 0.", data()->hash());
 
+  if ((torrentSize + chunkSize - 1) / chunkSize > std::numeric_limits<Bitfield::size_type>::max())
+    throw input_error("Torrent has more chunks than the chunk index can address.");
```
rakshasa, line 355: This can overflow `torrentSize`, need change how you calculate it.

xirvik: done

PR #976, opened 2026-09-19
Seed udns query ids from the system entropy source and require the QR flag

xirvik:
Query ids came from tv_usec and a query was not told apart from a reply.


rakshasa:
Could you merge latest master for all branches, it will make CI tests easier.

xirvik:
> Could you merge latest master for all branches, it will make CI tests easier.

Done

PR #977, opened 2026-09-19
Use a large state engine for UDP tracker transaction ids

xirvik:
A 31 bit engine seeded with 32 bits let one observed id predict every later id.


src/tracker/udp_router.cc
```diff
@@ -25,13 +25,20 @@
...
   std::random_device rd;
-  std::mt19937       mt(rd());
+  std::seed_seq      seed{rd(), rd(), rd(), rd(), rd(), rd(), rd(), rd()};
+
+  random_engine engine;
+  engine.seed(seed);
 
-  m_random_engine.seed(mt());
+  return engine;
+}
```
rakshasa, line 39: This should be an anonymous namespace function.

PR #978, opened 2026-09-19
Reject loopback, link-local and private addresses from remote peer lists

xirvik:
Any party could put an internal address in a peer list and have it connected to.


src/torrent/net/socket_address.cc
```diff
@@ -129,7 +129,7 @@ sa_is_loopback(const sockaddr* sa) {
   case AF_INET6:
     if (sa_is_v4mapped(sa)) {
       uint32_t v4_raw = sin6_addr32_index(reinterpret_cast<const sockaddr_in6*>(sa), 3);
-      return (ntohl(v4_raw) >> 24) == 127;
+      return (v4_raw >> 24) == 127;
```
rakshasa: This change looks wrong.

Also, we can just get the byte at the sockaddr_in6 sin6_addr index for this check.

xirvik: > This change looks wrong.
> 
> Also, we can just get the byte at the sockaddr_in6 sin6_addr index for this check.

Better?

PR #980, opened 2026-09-19
Cap the number of info hashes the DHT keeps peer lists for

xirvik:
Entries are pruned only when empty, so remote announces could grow the table freely.


src/dht/dht_server.cc
```diff
@@ -384,6 +384,9 @@ DhtServer::create_announce_peer_response(const DhtMessage& req, const sockaddr*
 
   DhtTracker* tracker = m_router->get_tracker(*HashString::cast_from(info_hash.data()), true);
 
+  if (tracker == NULL)
+    throw dht_error(dht_error_generic, "Tracking too many info hashes");
```
rakshasa, line 388: This seems wrong, you also need to prune old inactive trackers.

rakshasa:
This needs to periodically prune stale trackers, e.g. 5% every hour if at max limit.

xirvik:
> This needs to periodically prune stale trackers, e.g. 5% every hour if at max limit.

Done

src/dht/dht_router.h
```diff
@@ -134,6 +145,7 @@ class DhtRouter : public DhtNode {
   std::optional<std::deque<contact_t>> m_contacts;
 
   int                 m_numRefresh{0};
+  uint32_t            m_lastTrackerEvict{0};
```
rakshasa, line 148: std::chrono::seconds

src/dht/dht_router.cc
```diff
@@ -479,13 +479,45 @@ DhtRouter::receive_timeout() {
...
   m_server.update();
 
   m_numRefresh++;
 }
 
+// Announces that keep every entry fresh would otherwise lock new info hashes
+// out of a full table, so make room by dropping the stalest entries.
+void
+DhtRouter::evict_stale_trackers() {
+  auto now = this_thread::cached_seconds().count();
```
rakshasa, line 495: It's a thread local variable, just call the function in-line.

src/dht/dht_router.cc
```diff
@@ -479,13 +479,45 @@ DhtRouter::receive_timeout() {
...
+    return;
+
+  std::vector<std::pair<uint32_t, HashString>> trackers;
+  trackers.reserve(m_trackers.size());
+
+  for (const auto& [hash, tracker] : m_trackers)
+    trackers.emplace_back(tracker->last_seen(), hash);
+
+  auto last = trackers.begin() + num_tracker_evict;
+  std::nth_element(trackers.begin(), last, trackers.end(), [](const auto& a, const auto& b) { return a.first < b.first; });
```
rakshasa, line 507: There's no need to copy the whole vector, just populate an eviction vector you keep sorted.

```
    for (const auto& item : source) {
        if (target.size() < maxSize) {
            // Vector isn't full yet: insert and maintain heap property
            target.push_back(item);
            std::push_heap(target.begin(), target.end(), compareByTimestamp);
        } 
        // O(1) Check: Is the new element older than the current oldest "new" element?
        else if (item.timestamp < target.front().timestamp) {
            // Remove the element with the largest timestamp
            std::pop_heap(target.begin(), target.end(), compareByTimestamp);
            target.back() = item; // Overwrite the evicted element
            
            // Push the new element into the heap
            std::push_heap(target.begin(), target.end(), compareByTimestamp);
        }
    }
```

PR #984, opened 2026-09-20
TrackerHttp: don't clobber m_current_family when there is no next family to try

axolotty:
*Reworked after review: the original patch touched `TrackerUdp`, which turned out not to
be the cause. See the comment below for what the actual bug was.*

## Problem

Tracker errors are reported with the wrong address family whenever only one family is
tried. A request sent over IPv4 is reported as `v6`:

```
tracker_http: sending event   : state:started family:AF_INET url:http://example.com/announce
tracker_http: received reply  : state:started url:http://example.com/announce
tracker_http: received failure: url:http://example.com/announce : v6 : Could not parse bencoded data: ...
```

`TrackerHttp::send_next_family()` advances `m_current_family = m_next_family` before
checking whether there is anything to advance to. When `m_next_family` is `AF_UNSPEC` it
returns `false`, but `m_current_family` has already been overwritten with `AF_UNSPEC`.
`receive_failed()` then calls `generate_error_message(m_current_family, ...)` on that
clobbered value, and the label is picked with
`current_family == AF_INET ? "v4 : " : "v6 : "`, so anything that is not `AF_INET` prints
as `v6`.

The same happens in the second early return: `m_current_family` is set to a family that
is blocked and therefore never sent, and the failure of the family that *was* sent gets
labeled with it.

It is only wrong when a single family is used, which is why it is easy to miss: in the
two-family case the last attempt really is IPv6, so `AF_UNSPEC` lands on the same branch
and the label is right by accident.

Single-family requests happen whenever `request_families()` returns `AF_UNSPEC` as the
next family — with `network.block.ipv4`/`network.block.ipv6` set, or simply when the
announce URL uses a numeric IPv4 address, so no special configuration is required to hit
this.

## Fix

Only advance `m_current_family` once it is certain the request will be sent, so the
failure is labeled with the family actually used. `m_next_family` is still consumed in
every return path, and the behaviour is otherwise unchanged.

## Verification

Built at `v0.16.23` (the file is identical on master) and run in two containers that
differ only by this patch, same config, same torrent, announcing to a URL that returns
non-bencoded data:

| | `network.block.ipv6.set = yes` (one family) | unset (both families) |
|---|---|---|
| before | `v6 : Could not parse bencoded data: ...` ❌ | `v6 : Bind address ... \| v4 : Could not parse ...` ✅ |
| after | `v4 : Could not parse bencoded data: ...` ✅ | `v6 : Bind address ... \| v4 : Could not parse ...` ✅ |

In both cases the `tracker_requests` log shows `sending event ... family:AF_INET` for the
request that failed, so only the label changed. The two-family output is byte-identical
before and after.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01114UzppjAmKDQ1FYTmjVME


rakshasa:
This should be handled properly by `NetworkConfig::bind_address_for_connect_unsafe()` and `UdpRouter::updated_network_config()`.

Check `tracker_requests` log category for `udp-router` messages.

axolotty:
You were right, and following that pointer led me to the actual bug — so I've reworked
this PR rather than keep the `TrackerUdp` change.

Checking `tracker_requests` showed the request that failed was sent over IPv4:

```
sending event   : state:started family:AF_INET url:http://example.com/announce
received reply  : state:started url:http://example.com/announce
received failure: url:http://example.com/announce : v6 : Could not parse bencoded data: ...
```

So nothing was attempting IPv6 — the label was simply wrong. The cause is in
`TrackerHttp::send_next_family()`: it does `m_current_family = m_next_family` before
checking whether there is a next family, so on the `AF_UNSPEC` early return
`m_current_family` is already clobbered by the time `receive_failed()` passes it to
`generate_error_message()`, where `current_family == AF_INET ? "v4 : " : "v6 : "` turns
anything non-`AF_INET` into `v6`. Same for the blocked-family early return, which leaves
it set to a family that was never sent.

It only shows when a single family is used — in the two-family case the last attempt
really is IPv6, so the clobbered value lands on the same branch and the label is right by
accident. Besides the block options, `request_families()` also returns `AF_UNSPEC` as the
next family when the announce URL is a numeric IPv4 address, so it reproduces without any
special configuration.

On the UDP side you were right that there was nothing to fix: the router is closed for a
blocked family and `UdpRouter::connect()` returns early, and `TrackerUdp` passes the
family explicitly to `generate_error_message()` so it cannot mislabel. I dropped that
commit.

Verification is in the updated description (A/B build, same config and torrent, only the
label changes; two-family output byte-identical before and after).

